PR #38, opened 2023-12-01
Setup scripts for local dev

kavilla:
Setup scripts will assume it's in the root plugin directory of OSD and places the `scripts` folder into the OSD/scripts folder for run purposes with the following command:
```
yarn setup
```

This then will allow from the root OSD directory to run
```
yarn assistant:start
```
which will setup up the OpenSearch cluster as expected.

---
⚠️ Please note, at the time of writing this the `feature/os-assistant` and `feature/2.11/os-assistant` added the [feature](https://github.com/opensearch-project/OpenSearch-Dashboards/commit/e211cabc6be6e6f9f956f325a0df480976ffc50a) for installing OpenSearch assistant related plugins. Not main or 2.x.
---

### DIFF
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -41,7 +41,6 @@ Ultimately, your directory structure should look like this:
 │       └── dashboards-assistant
 ```
 
-
 ## Build
 
 To build the plugin's distributable zip simply run `yarn build`.
@@ -56,6 +55,35 @@ Example output: `./build/assistantDashboards-2.11.0.zip`
   - Starts OpenSearch-Dashboards and includes this plugin. OpenSearch-Dashboards will be available on `localhost:5601`.
   - Please run in the OpenSearch-Dashboards root directory
   - You must have OpenSearch running with the ML Commons plugin
+- Quick setup version utilizing Bedrock:
+
+1. Export keys locally:
+```sh
+# defaults to us-west-2
+# Requires permission to Bedrock and granted access to Claude
+export REGION=
+export AWS_ACCESS_KEY_ID=
+export AWS_SECRET_ACCESS_KEY=
+export AWS_SESSION_TOKEN=
+```
+2. Setup OSD plugins
+```sh
+# from OSD root folder
+cd plugins
+git clone https://github.com/opensearch-project/dashboards-assistant
+(cd dashboards-assistant && yarn setup)
+
+# example with security plugin
+git clone --depth 1 --branch $VERSION https://github.com/opensearch-project/security-dashboards-plugin.git
+
+cd ../
+yarn osd bootstrap
+```
+2. Start up an OpenSearch snapshot with ML-Commons, SQL, Observability, Alerting, and Security. Adds some default models. Finally, starts up OpenSearch Dashboards:
+```sh
+# from OSD root folder
+yarn start:assistant
+```
 
 ## Test
 
--- a/scripts/assistant/add_model.sh
+++ b/scripts/assistant/add_model.sh
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+set -e
+
+function add_model() {
+  [ -z "$SESSION_TOKEN" ] && SESSION_TOKEN_VALUE="" || SESSION_TOKEN_VALUE='"session_token": "'$SESSION_TOKEN'"'
+  [ -z "$SESSION_TOKEN_VALUE" ] && echo "[ User session disabled ]"
+
+  ENDPOINT=https://${USERNAME}:${PASSWORD}@${BIND_ADDRESS}:${BIND_PORT}
+
+  curl -s -k "${ENDPOINT}/_cluster/settings" -XPUT -H 'Content-Type: application/json' -d '{
+    "persistent" : {
+      "plugins.ml_commons.trusted_connector_endpoints_regex":
+      [ "^https://runtime\\.sagemaker\\..*[a-z0-9-]\\.amazonaws\\.com/.*$",
+        "^https://api\\.openai\\.com/.*$",
+        "^https://api\\.cohere\\.ai/.*$",
+        "^https://bedrock-runtime\\.[a-z0-9-]+\\.amazonaws\\.com/.*$"
+      ]
+    }
+  }' | jq
+
+  # shellcheck disable=2016
+  CONNECTOR=$(curl -s -k "${ENDPOINT}/_plugins/_ml/connectors/_create" -XPOST -H 'Content-Type: application/json' -d '{
+    "name": "BedRock test claude Connector",
+    "description": "The connector to BedRock service for claude model",
+    "version": 1,
+    "protocol": "aws_sigv4",
+    "parameters": {
+        "region": "'$REGION'",
+        "service_name": "bedrock",
+        "anthropic_version": "bedrock-2023-05-31",
+        "endpoint": "bedrock.'$REGION'.amazonaws.com",
+        "auth": "Sig_V4",
+        "content_type": "application/json"
+    },
+    "credential": {
+      "access_key": "'$ACCESS_KEY_ID'",
+      "secret_key": "'$SECRET_ACCESS_KEY'",
+      '"$SESSION_TOKEN_VALUE"'
+    },
+    "actions": [
+      {
+        "action_type": "predict",
+        "method": "POST",
+        "url": "https://bedrock-runtime.'"$REGION"'.amazonaws.com/model/anthropic.claude-v1/invoke",
+        "headers": {
+          "content-type": "application/json",
+          "x-amz-content-sha256": "required"
+        },
+        "request_body": "{\"prompt\":\"${parameters.prompt}\", \"max_tokens_to_sample\":${parameters.max_tokens_to_sample}, \"temperature\":${parameters.temperature},  \"anthropic_version\":\"${parameters.anthropic_version}\" }"
+      }
+    ]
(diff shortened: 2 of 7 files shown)


public/tabs/history/chat_history_page.tsx
```diff
@@ -20,7 +20,7 @@ import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { FormattedMessage } from '@osd/i18n/react';
 import { useDebounce, useObservable } from 'react-use';
 import cs from 'classnames';
-import { TAB_ID } from '../../utils/constants';
+import { TAB_ID } from '../../utils';
```
kavilla, line 23: i was getting errors with this before so changed it to this.

scripts/assistant/start.sh
```diff
@@ -0,0 +1,105 @@
...
+CWD=$(pwd)
+SNAPSHOT_DIR="$CWD/.opensearch"
+LOGS_DIR="$SNAPSHOT_DIR/$PACKAGE_VERSION/logs"
+
+# Main function
+function execute() {
+  export initialAdminPassword=$PASSWORD
+  CLUSTER_SETTINGS="snapshot --assistant --security"
+  CLUSTER_SETTINGS+=" -E plugins.ml_commons.only_run_on_ml_node=true"
+  CLUSTER_SETTINGS+=" -E plugins.ml_commons.memory_feature_enabled=true"
```
kavilla, line 87: @xluo-aws do you know if there are more configurations I need to set for OpenSearch?

scripts/pre_install.sh
SuZhou-Joe, line 1: One question: From the code it copies the scripts to OSD and change the version field of OSD's package.json. But do we have to cp files to OSD? Is there any limit to execute the script in the dashboards-assistant home directory?

scripts/assistant/start.sh
xluo-aws: Need to add this OSD configuration "assistant.chat.enabled: true" otherwise the chatbot icon and box will not be shown on the top right

wanglam: Do we need to set `only_run_on_ml_node` to `false`? The model must be run in ml node if the setting is `true`.

scripts/assistant/start.sh
```diff
@@ -0,0 +1,105 @@
...
+  run_opensearch || clean
+  check_opensearch_status
+  echo "[ Attempting to add models... ]"
+  (add_model > $LOGS_DIR/add_model.log 2>&1 || clean) & 
+  echo "Results found in $LOGS_DIR/add_model.log"
+
+  export OPENSEARCH_USERNAME=kibanaserver
+  export OPENSEARCH_PASSWORD=kibanaserver
+  echo "[ Starting OpenSearch Dashboards... ]"
+  OSD_SETTINGS="--dev --security --assistant.chat.enabled=true"
```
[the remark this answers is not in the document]
xluo-aws: "persistent" : {
    
    "plugins.ml_commons.only_run_on_ml_node":"false",
    "plugins.ml_commons.memory_feature_enabled":"true"
  }

scripts/assistant/add_model.sh
```diff
@@ -0,0 +1,88 @@
...
+    "credential": {
+      "access_key": "'$ACCESS_KEY_ID'",
+      "secret_key": "'$SECRET_ACCESS_KEY'",
+      '"$SESSION_TOKEN_VALUE"'
+    },
+    "actions": [
+      {
+        "action_type": "predict",
+        "method": "POST",
+        "url": "https://bedrock-runtime.'"$REGION"'.amazonaws.com/model/anthropic.claude-v1/invoke",
```
ruanyl, line 44: Shall we use claude instant by default? @anirudha mentioned claude instant is the long term direction.

ruanyl (commented):
Maybe for future improvement, we could add setup guides for, for example, openai, sagemaker, etc.

scripts/assistant/add_model.sh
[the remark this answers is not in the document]
kavilla: do we have that available? totally will prefer that.

DEVELOPER_GUIDE.md
```diff
@@ -24,6 +24,35 @@ Ultimately, your directory structure should look like this:
...
+```sh
+# from OSD root folder
+cd plugins
+git clone https://github.com/opensearch-project/dashboards-assistant
+(cd dashboards-assistant && yarn setup)
+
+# example with security plugin
+git clone --depth 1 --branch $VERSION https://github.com/opensearch-project/security-dashboards-plugin.git
+
+cd ../../
```
joshuali925, line 47: you are in `plugins` directory, `cd ../..` goes to parent of osd dir instead of osd dir

DEVELOPER_GUIDE.md
```diff
@@ -24,6 +24,35 @@ Ultimately, your directory structure should look like this:
...
+yarn osd bootstrap
+```
+2. Start local environment
+```sh
+# from OSD root folder
+yarn start:assistant
```
joshuali925, line 53: i understand that this makes development easier, but would a plugin specific start command make it less transparent to what's going on? e.g. how does developers tell the difference between `yarn start` and `yarn start:assistant`

maybe add a comment to code reference on what this does?

scripts/assistant/add_model.sh
```diff
@@ -0,0 +1,88 @@
...
+        "service_name": "bedrock",
+        "anthropic_version": "bedrock-2023-05-31",
+        "endpoint": "bedrock.'$REGION'.amazonaws.com",
+        "auth": "Sig_V4",
+        "content_type": "application/json"
+    },
+    "credential": {
+      "access_key": "'$ACCESS_KEY_ID'",
+      "secret_key": "'$SECRET_ACCESS_KEY'",
+      '"$SESSION_TOKEN_VALUE"'
```
joshuali925, line 38: i think this will fail, if `$SESSION_TOKEN_VALUE` is empty, then there's an extra comma after `secret_key`. `jq` or include the comma in the variable might help, or add a TODO and i'll fix this later

scripts/assistant/start.sh
```diff
@@ -0,0 +1,105 @@
...
+  OSD_SETTINGS+=" --home.newHomepage=true"
+  eval "$CWD/scripts/use_node $CWD/scripts/opensearch_dashboards $OSD_SETTINGS" || clean
+}
+
+execute
+clean
+exit 0
```
joshuali925, line 105: why `exit 0` here? if exit status is not `0`, the script should exit before it reaches this line because you have `set -e`

joshuali925 (commented):
thanks for adding this, a few comments

scripts/assistant/add_model.sh
kavilla: yeah, i was thinking about that and verified that that it doesn't cause any issue would occur with sending a request with an unneeded  `,` i didn't run into any issue but it could be OS specific. do you think we should add something there or prefer the prepend to the variable ?

scripts/assistant/start.sh
kavilla: yeah.  within OSD i had intent of start up verification tests. CI can verify exit 0. but since the clean is the last method which outputs the process ID. the output would be non-`1`. can remove but might be worth incase we would like to setup a CI using a quick setup script.

DEVELOPER_GUIDE.md
kavilla: good catch!

DEVELOPER_GUIDE.md
```diff
@@ -24,6 +24,35 @@ Ultimately, your directory structure should look like this:
...
 ```
 
+Setup local stack
+
+1. Export keys locally:
+```sh
+# defaults to us-west-2
+export REGION=
+export AWS_ACCESS_KEY_ID=
+export AWS_SECRET_ACCESS_KEY=
```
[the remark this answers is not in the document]
kavilla: good call added

DEVELOPER_GUIDE.md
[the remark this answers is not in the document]
kavilla: good call added

scripts/assistant/add_model.sh
kavilla: apologies, not sure what to add here or change.  this models script was the one i got so def willing to change it to the correct value 😅

scripts/pre_install.sh
kavilla: i was having some limitation with the node executing within this repo. this limitation might be relatively recently but have digged into deep enough

DEVELOPER_GUIDE.md
kavilla: good call. added a comment. this outputs to the terminal each step but it really is different than `start` so the docs could be confusing when reading the difference.

scripts/assistant/add_model.sh
joshuali925: hm standard json parser shouldn't allow this, but if it's working fine in opensearch then it's ok

scripts/assistant/start.sh
joshuali925: `clean` has `kill || true`, it always exits 0. i don't think i really understood the comment but if you think it's needed you can leave it

joshuali925 (approved):
thanks

scripts/assistant/start.sh
kavilla: ah shoot, good catch that might actually will clean it up. however, if other folks on this PR thinks this ok I can clean it up and a quick follow up to avoid re-approval

PR #41, opened 2023-12-05
display feedback state when loading a past conversation

raintygao:
### Description
display feedback state when loading a past conversation

### Issues Resolved

### DIFF
--- a/common/types/chat_saved_object_attributes.ts
+++ b/common/types/chat_saved_object_attributes.ts
@@ -12,7 +12,7 @@ export interface Interaction {
   conversation_id: string;
   interaction_id: string;
   create_time: string;
-  additional_info?: Record<string, Record<string, boolean>>;
+  additional_info?: Record<string, string | Record<string, boolean>>;
   parent_interaction_id?: string;
 }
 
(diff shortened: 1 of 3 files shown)


wanglam (approved):

common/types/chat_saved_object_attributes.ts
```diff
@@ -12,7 +12,7 @@ export interface Interaction {
   conversation_id: string;
   interaction_id: string;
   create_time: string;
-  additional_info?: Record<string, unknown>;
```
SuZhou-Joe, line 15: The key-value pair does not have to be Record<string, boolean> I suppose?

raintygao: Yes, it may can also be `string`. But i assumed this could be updated when needed in future or predefined.

SuZhou-Joe: image
What about `{ feedback?: SendFeedbackBody, [key: string]: unknown }`.

My intent to set value as `unknown` is letting consumers to cast to other types explicitly in their functions if they want to consume specific field and know exactly what the type those values are.

raintygao: That makes sense.

SuZhou-Joe (approved):

PR #52, opened 2023-12-07
consider security disabled case in permission check

joshuali925:
### Description
see #9 
rebased #9 to agent framework branch

### Issues Resolved

### DIFF
--- a/public/plugin.tsx
+++ b/public/plugin.tsx
@@ -12,8 +12,8 @@ import {
 import { createGetterSetter } from '../../../src/plugins/opensearch_dashboards_utils/common';
 import { HeaderChatButton } from './chat_header_button';
 import { AssistantServices } from './contexts/core_context';
-import { SessionsService } from './services/sessions_service';
 import { SessionLoadService } from './services/session_load_service';
+import { SessionsService } from './services/sessions_service';
 import {
   ActionExecutor,
   AppPluginStartDependencies,
@@ -34,7 +34,7 @@ interface PublicConfig {
 }
 
 interface UserAccountResponse {
-  data: { roles: string[]; user_name: string; user_requested_tenant: string | null };
+  data: { roles: string[]; user_name: string; user_requested_tenant?: string };
 }
 
 export class AssistantPlugin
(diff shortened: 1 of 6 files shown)


public/plugin.tsx
```diff
@@ -46,23 +51,28 @@ export class AssistantPlugin
...
       return async () => {
-        if (enabled === undefined) {
-          const account = await getAccount();
-          enabled = account.data.roles.some((role) =>
-            ['all_access', 'assistant_user'].includes(role)
-          );
+        if (setupDeps.securityDashboards === undefined)
+          return { data: { roles: ['all_access'], user_name: 'dashboards_user' } };
+        if (account === undefined) {
+          account = await core.http
```
wanglam, line 64: Seems the `account` was not be modified in other place, shall we change to `const account`?

public/plugin.tsx
```diff
@@ -46,23 +51,28 @@ export class AssistantPlugin
...
+      let account: UserAccountResponse;
       return async () => {
-        if (enabled === undefined) {
-          const account = await getAccount();
-          enabled = account.data.roles.some((role) =>
-            ['all_access', 'assistant_user'].includes(role)
-          );
+        if (setupDeps.securityDashboards === undefined)
+          return { data: { roles: ['all_access'], user_name: 'dashboards_user' } };
+        if (account === undefined) {
```
wanglam, line 63: The `account` is always be undefined here, shall we remove this check?

public/plugin.tsx
```diff
@@ -108,7 +116,8 @@ export class AssistantPlugin
           console.warn(`Action executor type ${actionType} is already registered.`);
         actionExecutors[actionType] = execute;
       },
-      assistantEnabled,
+      chatEnabled: () => this.config.chat.enabled,
+      userHasAccess: async () => await getAccount().then(checkAccess),
```
wanglam, line 120: The result of `userHasAcess` may not the same as the `HeaderChatButton` one. Shall we use the same getAccount result?

public/plugin.tsx
joshuali925: it's a closure to save http calls, i think it's acceptable that if user permission changes, UI needs a refresh to see the updated results. the real permission check happens on backend anyways

public/plugin.tsx
joshuali925: not on the second `getAccount()` call

public/plugin.tsx
joshuali925: they should be the same, account is cached

wanglam (approved):

raintygao (approved):

PR #53, opened 2023-12-08
[Incontext Insights] wrapper component and service

kavilla:
### Description
Registry and component to be used by plugins.

issue resolves: https://github.com/opensearch-project/dashboards-assistant/issues/74

Example of usage:
https://github.com/opensearch-project/OpenSearch-Dashboards/commit/25d705683ba7f6b802344343083669cb66f19926
https://github.com/opensearch-project/alerting-dashboards-plugin/pull/852

Screenshot 2023-12-13 at 5 22 11 PM

Screenshot 2023-12-13 at 5 30 10 PM

### Issues Resolved


public/plugin.tsx
```diff
@@ -110,13 +140,17 @@ export class AssistantPlugin
       },
       assistantEnabled,
       assistantActions,
+      registerPalantir: this.palantirRegistry!.register.bind(this.palantirRegistry),
+      getPalantir: this.palantirRegistry!.get.bind(this.palantirRegistry),
```
wanglam, line 144: Is this `!` necessary? Seems `this.palantirRegistry` is initialized in top of setup method. It's not undefined here.

public/plugin.tsx
```diff
@@ -110,13 +140,17 @@ export class AssistantPlugin
...
     };
   }
 
-  public start(core: CoreStart, startDeps: AppPluginStartDependencies): AssistantStart {
+  public start(core: CoreStart): AssistantStart {
     setCoreStart(core);
 
-    return {};
+    return {
+      getPalantir: this.palantirRegistry!.get.bind(this.palantirRegistry),
```
wanglam, line 152: Prefer add a not undefined check for `this.palantirRegistry`. We can throw an error if it's undefined.

public/chat_header_button.tsx
```diff
@@ -140,6 +142,26 @@ export const HeaderChatButton: React.FC<HeaderChatButtonProps> = (props) => {
     };
   }, []);
 
+  useEffect(() => {
```
wanglam, line 145: Shall we add some unit tests for these logic?

public/services/palantir/palantir_registry.ts
```diff
@@ -0,0 +1,58 @@
...
+      item.forEach((palantir: Palantir) =>
+        this.registry.set(palantir.key, {
+          key: palantir.key,
+          description: palantir.description,
+          suggestion: palantir.suggestion,
+        })
+      );
+    } else {
+      const palantir = item as Palantir;
+      this.registry.set(palantir.key, palantir);
```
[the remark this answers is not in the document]
kavilla: sorry could you provide an example?

public/chat_header_button.tsx
kavilla: definitely. wanted to get this out for feedback.

public/services/palantir/palantir_registry.ts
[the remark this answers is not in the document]
wanglam: Sure. Just wondering if we need to refactor like this:
```suggestion
      this.registry.set(palantir.key, {
          key: palantir.key,
          description: palantir.description,
          suggestion: palantir.suggestion,
        });
```
It would be consistent with above.

config.ts
```diff
@@ -0,0 +1,15 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { schema, TypeOf } from '@osd/config-schema';
+
+export const configSchema = schema.object({
+  enabled: schema.boolean({ defaultValue: true }),
+  chat: schema.object({
+    enabled: schema.boolean({ defaultValue: true }),
+  }),
```
SuZhou-Joe, line 12: Do we need two `enabled` flag?

public/components/palantir/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+    <EuiPopover
+      key={key ?? palantirInput.key}
+      button={anchorContent}
+      isOpen={isVisible}
+      closePopover={closeAssistantPopover}
+      anchorClassName="palantirAnchor"
+      anchorPosition="upCenter"
+      display="block"
+    >
+      <EuiPopoverTitle>{children}</EuiPopoverTitle>
```
ruanyl, line 90: Q: Why render `children` again in popover title?

public/components/palantir/index.tsx
ruanyl, line 1: Just thinking if that's a good idea to implement `Palantir` as a generic component/service in OSD core? Instead of having core plugins or other plugins to depend on `dashboards-assistant` which doesn't sound optimum to me. Also having it in `dashboards-assistant` to serve a specific use case makes it hard to extend, as assistant might NOT be the only use case(Maybe a bit over thinking?).

Currently, the only part that makes `Palantir` component "assistant-specific" is the `EuiPopoverFooter` section which renders the action(s) to execute. I can move `Palantir` to OSD core and let the plugins to register the footer actions, here are some quick dummy code:

```typescript
interface ActionConfig {
  type: string
  render: (palantirInput: PalantirInput, onSubmit: () => void) => React.ReactNode
}

type ActionListener = (palantirInput: PalantirInput) => void

// PalantirService in core
class PalantirService {
  actions = new Map<string, ActionConfig>
  handlers = new Set<ActionListener>()

  registerAction(actionConfig: ActionConfig) {
    this.actions.set(actionConfig.type, actionConfig)
  }

  addActionHandler(listener: ActionListener) {
    this.handlers.add(listener)
  }

  removeActionHandler(listener: ActionListener) {
    this.handlers.delete(listener)
  }

  handlePalantir(palantirInput: PalantirInput) {
    this.handlers.forEach(fn => fn(palantirInput))
  }
}
```

Then the assistant plugin register an action
```typescript
core.palantirService.registerAction({
  type: 'assistant',
  render: (palantirInput, onSubmit) =>
    <>
      <EuiText size="xs">Continue in chat</EuiText>
      <EuiBadge
        color="hollow"
        iconType="chatRight"
        iconSide="left"
        onClick={onSubmit}
        onClickAriaLabel="Click for suggestion"
      >
        {palantirInput.suggestion}
      </EuiBadge>
    </>
})
```

Move the current Palantir component to core and change the EuiPopoverFooter to:
```typescript
      <EuiPopoverFooter>
      {palantirService.actions.get(targetAction)?.render(palantirInput, () => onSubmitClick(palantirInput))}
      </EuiPopoverFooter>
```

And this is how Palantir will be used, `targetAction` will be specified so that Palantir know what action to display
```typescript
<Palantir targetAction="assistant" title="xx" input={input}><WhateverComponent /></Palantir>
```

And assistant plugin handle the action:
```typescript
  useEffect(() => {
    const handleSuggestion = (event: { suggestion: string }) => {
      // open chat window
      setFlyoutVisible(true);
      // start a new chat
      props.assistantActions.loadChat();
      // send message
      props.assistantActions.send({
        type: 'input',
        contentType: 'text',
        content: event.suggestion,
        context: { appId },
      });
    };
    core.palantirService.addActionHandler(handleSuggestion);
    return () => {
      core.palantirService.removeActionHandler(handleSuggestion);
    };
  }, []);
```

config.ts
kavilla: `plugin_id.enabled` is actually default behavior to "install" itself when the plugin discovery process is occurring. the current `chat.enabled` here will still make all this code still be available in the browser but it doesn't do the setup steps.

however, by adding this enabled. the plugin will not be discovered and hosted in the browser. if the chat is not enabled and the expected behavior is none of the code gets added to the browser then we should opt for removing the `chat.enabled`

config.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { schema, TypeOf } from '@osd/config-schema';
+
+export const configSchema = schema.object({
+  enabled: schema.boolean({ defaultValue: true }),
```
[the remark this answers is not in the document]
kavilla: for distribution?

we could add that to the bundle: https://github.com/opensearch-project/opensearch-build/blob/main/config/opensearch_dashboards-default.x.yml#L199

and then development we can avoid a specific starting up dev instructions since it would be a little bit confusing to clone this repo and it not to be available out of the box.

public/chat_header_button.tsx
```diff
@@ -140,6 +142,26 @@ export const HeaderChatButton: React.FC<HeaderChatButtonProps> = (props) => {
...
+        contentType: 'text',
+        content: event.suggestion,
+        context: { appId },
+      });
+    };
+    registry.on('onSuggestion', handleSuggestion);
+    return () => {
+      registry.off('onSuggestion', handleSuggestion);
+    };
+  });
```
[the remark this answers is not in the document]
kavilla: apologies i didn't verify is the chat header button being re-rendered multiple times? if so then i think we should probably refactor this component

public/components/palantir/index.tsx
kavilla: i do like the way you think I was original planning on that path but thinking how far do we want to go.

there is a lot of existing services that are available within core OSD but are not used since they require proper wiring up.

i think eventual idea here would be to house this service and component within this plugin for the sake avoiding any breaking changes or issues within core OSD as this is still experimental. once enough feedback has been collected and the component is in its final state. then the core structure of this component can be placed into core OSD (and probably OUI). To which then this plugin then extends that core functionality of which you have mocked up and then if plugin developers decide that they want to hook up their features we can provide a guide but if they  want to use a default functionality then they can just keep to using this plugin.

thoughts?

cc: @lezzago

public/components/palantir/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  key?: string;
+  input?: PalantirInput;
+  onSubmit?: () => void;
+}) => {
+  const registry = getPalantirRegistry();
+  const [isVisible, setIsVisible] = useState(false);
+
+  // TODO: error handling for bad implemention from plugins or just don't render?
+  const onAnchorClick = () => setIsVisible((visible) => !visible);
+  const onSubmitClick = async (palantirInput: PalantirInput) => {
```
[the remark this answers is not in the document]
kavilla: good catch. yeah i can clean this up. as you can see this can is kind of showcasing the original though that plugin developers can provide their onSubmit function and originally i was implementing utilizing this plugin to call to pass in an onSubmit function that executing and at the time it was making an async execution.

public/components/palantir/index.tsx
kavilla: i can sanitize this as the HTML is still coming through but I believe it was to the original UX suggestion. but it has been a while.

cc: @opensearch-project/opensearch-ux

public/plugin.tsx
kavilla: good catch will addrses.

public/services/palantir/palantir_registry.ts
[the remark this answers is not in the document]
kavilla: ah i see what you mean. do you think i should take it one step further and make a helper function to map it it both places?

config.ts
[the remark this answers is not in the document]
ruanyl: Makes sense, thanks for clarification!

public/chat_header_button.tsx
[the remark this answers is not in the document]
ruanyl: I was thinking maybe add an empty dependency array `}, []);` if more robust if the event is supposed to be subscribed once when the component is mounted.

public/components/palantir/index.tsx
ruanyl: I see, I was thinking the `children` could be something else other than a simple text node, if that's the case, the popover may look a bit strange?

public/components/palantir/index.tsx
ruanyl: Totally understand your concern, I believe we do want the service we added to core to be stable and to serve more general use cases instead of specific one.

Just one question came in to my mind: if OSD core or other plugins which `import` `Palantir` component from `dashboards-assistant`, will that break the build process when `dashboards-assistant` is not in the plugins folder?

public/services/palantir/palantir_registry.ts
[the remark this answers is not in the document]
wanglam: Sure. I think that will be reasonable.

public/chat_header_button.tsx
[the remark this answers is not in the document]
kavilla: that's true will do.

public/components/palantir/index.tsx
kavilla: yeah. have updated it. also i think there were mocks at one point that add visualizations so its prolly a little bit over kill with the updated mocks i have modify it a bit

public/components/palantir/index.tsx
kavilla: > Totally understand your concern, I believe we do want the service we added to core to be stable and to serve more general use cases instead of specific one.

Totally. It would make it easier in some ways. Also we give us time to create a provider for the React context so that externally we can call certain functions from the assistant.

> Just one question came in to my mind: if OSD core or other plugins which import Palantir component from dashboards-assistant, will that break the build process when dashboards-assistant is not in the plugins folder?

Great question. So here's an example: https://github.com/opensearch-project/alerting-dashboards-plugin/pull/852/files there is a conditional there for actually using the component. and likely my fast follow will be handling the importing from a non existing file. probably an alias to trick the compiler

public/components/palantir/index.tsx
kavilla: here is out it currently looks. still not in its final form but without the children:

default:
Screenshot 2023-12-21 at 7 40 10 AM

hover:
Screenshot 2023-12-21 at 7 40 26 AM

clicked:
Screenshot 2023-12-21 at 7 40 37 AM

still need to keep improving it.

public/components/palantir/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+      {suggestionsPopoverFooter}
+    </>
+  );
+
+  const getPopoverBody = () => {
+    switch (palantirInput.type) {
+      case 'suggestions':
+        return suggestionsPopoverFooter;
+      case 'generate':
+        return generatePopoverBody;
```
kavilla, line 160: Screenshot 2024-01-02 at 3 56 58 PM

public/components/palantir/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+      case 'summary':
+        return summaryPopoverBody;
```
kavilla, line 162: Screenshot 2024-01-02 at 3 57 26 PM

public/components/palantir/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+      case 'summaryWithSuggestions':
+        return summaryWithSuggestionsPopoverBody;
```
kavilla, line 164: Screenshot 2024-01-02 at 3 56 28 PM

public/components/palantir/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+      case 'chat':
+        return chatPopoverBody;
```
kavilla, line 166: Screenshot 2024-01-02 at 4 03 49 PM

public/components/palantir/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+      case 'chatWithSuggestions':
+        return chatWithSuggestionsPopoverBody;
```
kavilla, line 168: Screenshot 2024-01-02 at 4 04 14 PM

kgcreative: For these, can we use the compressed version of form elements?

kavilla:
Updated screenshots:

View on load for 1 second:
Screenshot 2024-02-05 at 11 53 47 PM

Default view:
Screenshot 2024-02-05 at 11 52 17 PM

Hover over state:
Screenshot 2024-02-05 at 11 52 30 PM

On anchor click:
Screenshot 2024-02-05 at 11 52 37 PM

On seed suggestion click takes user to chat and closes popover:
Screenshot 2024-02-05 at 11 53 06 PM

On anchor click after new suggestion with seed context as root:
Screenshot 2024-02-05 at 11 53 26 PM

Example of hover state over h3 element:
Screenshot 2024-02-05 at 11 54 01 PM

config.ts
```diff
@@ -0,0 +1,21 @@
...
+  chat: schema.object({
+    enabled: schema.boolean({ defaultValue: false }),
+    rootAgentName: schema.conditional(
+      schema.siblingRef('enabled'),
+      true,
+      schema.string(),
+      schema.maybe(schema.string())
+    ),
```
SuZhou-Joe, line 17: This config is not needed anymore.

public/components/incontext_insight/index.tsx
```diff
@@ -0,0 +1,336 @@
...
+          incontextInsightAnchorIconClassList.remove('hoverEffect75');
+          setTimeout(() => {
+            incontextInsightAnchorIconClassList.remove('hoverEffect50');
+            setTimeout(() => {
+              incontextInsightAnchorIconClassList.remove('hoverEffect25');
+              incontextInsightAnchorButtonClassList.remove('hoverEffectUnderline');
+            }, 10);
+          }, 15);
+        }, 20);
+      }, 1125);
```
SuZhou-Joe, line 78: Can we make it a css animation?

config.ts
kavilla: addressed

public/components/incontext_insight/index.tsx
kavilla: I want to this code feels hacky and im willing to remove. this is basically only for the onload state and to avoid having the icon on hover push other elements while also be adding itself to the anchor and not causing issues to anchor as well before displaying block.

this is the biggest factor to decide whether we should allow this icon to show on any element

kavilla: i have cleaned up the code to make it a little bit cleaner

public/components/incontext_insight/index.scss
```diff
@@ -0,0 +1,144 @@
+// TODO: fix the styling due to not wanting the chat icon to push elements
+// Might be worth reconsidering the icon for absolute positioning
+.hoverEffectUnderline {
+  border: solid 2px transparent !important;
+  margin: -2px -5px -5px -7px !important;
+}
+
+.hoverEffect0 {
```
SuZhou-Joe, line 8: I am afraid this class name may be too generic, can we add a prefix like insight-context?

kavilla: good call
added.

common/types/config.ts
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { schema, TypeOf } from '@osd/config-schema';
+
+export const configSchema = schema.object({
+  enabled: schema.boolean({ defaultValue: true }),
```
wanglam, line 9: Since our plugin was experimental, shall we set this default value to `false`?
It seems this `enabled` was new added, shall we add to the `opensearch-build` repo?

kavilla: your call! imo i think true is fine. because this is the difference between the files not being loaded into the server or not. 

this i used to verify my code if not available if plugins still work with the component. and we can tell infra team if they want to fully disabled it and add it to the env variables. but for development it might be weird to install a plugin and configure your files to be true otherwise your files wont be loaded.

public/components/incontext_insight/index.tsx
```diff
@@ -0,0 +1,333 @@
...
+import { getIncontextInsightRegistry, getNotifications } from '../../services';
+// TODO: Replace with getChrome().logos.Chat.url
+import chatIcon from '../../assets/chat.svg';
+
+export interface IncontextInsightProps {
+  children?: React.ReactNode;
+}
+
+// TODO: add saved objects / config
+const container = document.createElement('div');
```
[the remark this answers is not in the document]
kavilla: Great catch. Left from a previous implementation where we dont need this anymore.

public/components/incontext_insight/index.tsx
```diff
@@ -0,0 +1,333 @@
...
+document.body.appendChild(container);
+
+export const IncontextInsight = ({ children }: IncontextInsightProps) => {
+  const anchor = useRef<HTMLDivElement>(null);
+  const [isVisible, setIsVisible] = useState(false);
+
+  useEffect(() => {
+    // TODO: use animation when not using display: none
+    if (anchor.current) {
+      const incontextInsightAnchorButtonClassList = anchor.current.parentElement?.querySelector(
```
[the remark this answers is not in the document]
kavilla: good question. feedback from ux.

to get the desired ux requested by ux team the actual anchor is not the button itself its the element being wrapped. so in the active state to get it point towards the child like this:

image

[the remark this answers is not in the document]
kavilla: so the element is being wrapped is the anchor and the wrapper div is the parent

wanglam (approved):
LGTM

SuZhou-Joe (approved):

public/components/incontext_insight/index.tsx
```diff
@@ -0,0 +1,333 @@
...
+        '.incontextInsightAnchorButton'
+      )?.classList;
+      const incontextInsightAnchorIconClassList = anchor.current.querySelector(
```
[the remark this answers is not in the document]
kavilla: good idea:

added fast follow here: https://github.com/opensearch-project/dashboards-assistant/issues/143

PR #54, opened 2023-12-08
Add tests for ChatPageContent component

ruanyl:
add MessageBubble unit tests
add MessageContent unit tests
remove MessageFooter which is not needed anymore

### Description
_Describe what this change achieves._

### Issues Resolved


public/tabs/chat/chat_page_content.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+        type: 'input',
+        content: 'what indices are in my cluster?',
+        contentType: 'text',
+      },
+      {
+        type: 'output',
+        content: 'here are the indices in your cluster: .kibana',
+        contentType: 'markdown',
+        suggestedActions: [{ actionType: 'send_as_input', message: 'suggested action mock' }],
+      },
```
wanglam, line 111: Shall we add more messages section? The test case is for only display suggestions for last output. We can check if the messages displayed in the previous output message.

ruanyl: updated

public/tabs/chat/chat_page_content.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+  return {
+    MessageBubble: ({ children }: { children?: React.ReactNode }) => (
+      <div aria-label="chat message bubble">{children}</div>
+    ),
+  };
+});
+
+jest.mock('./messages/message_content', () => {
+  return { MessageContent: () => <div /> };
+});
```
joshuali925, line 24: just curious why are child components mocked?

joshuali925 (approved):

SuZhou-Joe (approved):

public/tabs/chat/chat_page_content.test.tsx
ruanyl: That's a good question! There has been lots of discussions around this. For **unit test**, I tend to mock child components in RTL which are smart components, for example, the components that:
1. Access data via API call
2. Have side effects such as read/write local storage
3. Access context
4. Access global state

By mocking such components, I think it makes the tests concise and less fragile. And ofc, the child component should be covered by its own unit tests.

PR #59, opened 2023-12-11
Fix back button disappear on trace page

Hailong-am:
### Description
When click how was it generated twice, back button will disappear. Fix it by set chat window pre state when it changed.

image

### Issues Resolved


public/chat_header_button.tsx
```diff
@@ -65,7 +65,9 @@ export const HeaderChatButton: React.FC<HeaderChatButtonProps> = (props) => {
       selectedTabId,
       preSelectedTabId,
       setSelectedTabId: (tabId: TabId) => {
-        setPreSelectedTabId(selectedTabId);
+        if (selectedTabId !== tabId) {
```
wanglam, line 68: How about add this if check to `use_chat_actions`? We don't need to call setSelectedTabId if selectedTabId is already "trace".

wanglam (approved):

SuZhou-Joe (approved):

gaobinlong (approved):
LGTM.

PR #62, opened 2023-12-11
test: add unit test for save to notebook feature

raintygao:
### Description
1. add unit test for save to notebook feature
2. move toast logic to caller side

### Issues Resolved

### DIFF
--- a/public/components/notebook/notebook_name_modal.tsx
+++ b/public/components/notebook/notebook_name_modal.tsx
@@ -22,7 +22,7 @@ import { toMountPoint } from '../../../../../src/plugins/opensearch_dashboards_r
 export interface NotebookNameModalProps {
   onClose: () => void;
   // SaveChat hook depends on context. Runtime modal component can't get context, so saveChat needs to be passed in.
-  saveChat: (name: string) => void;
+  saveChat: (name: string) => Promise<string>;
 }
 
 export const NotebookNameModal = ({ onClose, saveChat }: NotebookNameModalProps) => {
@@ -42,15 +42,13 @@ export const NotebookNameModal = ({ onClose, saveChat }: NotebookNameModalProps)
 
       toasts.addSuccess({
         text: toMountPoint(
-          <>
-            <p>
-              This conversation was saved as{' '}
-              <EuiLink href={notebookLink} target="_blank">
-                {name}
-              </EuiLink>
-              .
-            </p>
-          </>
+          <p>
+            This conversation was saved as{' '}
+            <EuiLink href={notebookLink} target="_blank">
+              {name}
+            </EuiLink>
+            .
+          </p>
         ),
       });
     } catch (error) {
@@ -83,7 +81,7 @@ export const NotebookNameModal = ({ onClose, saveChat }: NotebookNameModalProps)
         </EuiModalBody>
 
         <EuiModalFooter>
-          <EuiButtonEmpty onClick={onClose} data-test-subj="confirmNotebookCancelButton">
+          <EuiButtonEmpty onClick={onClose} data-test-subj="cancelSaveToNotebookButton">
             Cancel
           </EuiButtonEmpty>
           <EuiButton
@@ -92,7 +90,7 @@ export const NotebookNameModal = ({ onClose, saveChat }: NotebookNameModalProps)
             isLoading={loading}
             disabled={name.length < 1}
             onClick={onSubmit}
-            data-test-subj="confirmNotebookConfirmButton"
+            data-test-subj="confirmSaveToNotebookButton"
           >
             Confirm name
           </EuiButton>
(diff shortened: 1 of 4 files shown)


SuZhou-Joe (approved):

public/components/notebook/notebook_name_modal.tsx
```diff
@@ -16,22 +16,54 @@ import {
...
+                {name}
+              </EuiLink>
+              .
+            </p>
+          </>
+        ),
+      });
+    } catch (error) {
+      if (error.message === 'Not Found') {
+        toasts.addDanger(
```
wanglam, line 58: Shall we add some unit tests for this Not Found error?

public/hooks/use_save_chat.tsx
wanglam, line 1: Shall we add separate test for this hook?

public/components/notebook/notebook_name_modal.tsx
```diff
@@ -16,22 +16,54 @@ import {
...
+    },
+  } = useCore();
   const [name, setName] = useState('');
   const [loading, setLoading] = useState(false);
 
   const onSubmit = useCallback(async () => {
     setLoading(true);
-    await saveChat(name);
+    try {
+      const notebookId = await saveChat(name);
```
ruanyl, line 40: The type of saveChat is `(name: string) => void;`, but here it expects it to be a promise which returns a `Promise<string>`.

public/components/notebook/notebook_name_modal.tsx
```diff
@@ -16,22 +16,54 @@ import {
...
+      const notebookLink = `./observability-notebooks#/${notebookId}?view=view_both`;
+
+      toasts.addSuccess({
+        text: toMountPoint(
+          <>
```
ruanyl, line 45: Seem you don't need to wrap it with `<></>`

public/components/notebook/notebook_name_modal.tsx
```diff
@@ -16,22 +16,54 @@ import {
...
+              <EuiLink href={notebookLink} target="_blank">
+                {name}
+              </EuiLink>
+              .
+            </p>
+          </>
+        ),
+      });
+    } catch (error) {
+      if (error.message === 'Not Found') {
```
ruanyl, line 57: This may out of the scope of this PR, but I'm wondering if it's ok to assume `observability plugin` is not installed when error message is `Not Found`? Because `Not Found` sounds like a general error message.

public/components/notebook/notebook_name_modal.tsx
```diff
@@ -42,18 +74,25 @@ export const NotebookNameModal = ({ onClose, saveChat }: Props) => {
...
+          <EuiButtonEmpty onClick={onClose} data-test-subj="confirmNotebookCancelButton">
+            Cancel
+          </EuiButtonEmpty>
           <EuiButton
             type="submit"
             fill
             isLoading={loading}
             disabled={name.length < 1}
             onClick={onSubmit}
+            data-test-subj="confirmNotebookConfirmButton"
```
ruanyl, line 95:
```suggestion
            data-test-subj="confirmSaveToNotebookButton"
```

public/components/notebook/notebook_name_modal.tsx
```diff
@@ -42,18 +74,25 @@ export const NotebookNameModal = ({ onClose, saveChat }: Props) => {
...
+              value={name}
+              onChange={(e) => setName(e.target.value)}
+              aria-label="Notebook name input"
+            />
           </EuiFormRow>
         </EuiModalBody>
 
         <EuiModalFooter>
-          <EuiButtonEmpty onClick={onClose}>Cancel</EuiButtonEmpty>
+          <EuiButtonEmpty onClick={onClose} data-test-subj="confirmNotebookCancelButton">
```
ruanyl, line 86:
```suggestion
          <EuiButtonEmpty onClick={onClose} data-test-subj="cancelSaveToNotebookButton">
```

public/hooks/use_save_chat.tsx
```diff
@@ -5,13 +5,10 @@
 
 import React from 'react';
```
ruanyl, line 6: Nice refactor! I guess then you can remove `import React from 'react';`

public/components/notebook/notebook_name_modal.tsx
raintygao: I think we can assume they are same because observability will register this route automatically if it is installed. OSD server will throw `Not Found` error if this route is not registered. I tried to import an error constant from OSD to avoid directly judge this string, but OSD doesn't export.

public/hooks/use_save_chat.tsx
raintygao: I tried to make most cases of this hook covered in the test of `notebook_name_modal`, i will create a new PR to cover the rest cases of this hook.

wanglam (approved):

PR #65, opened 2023-12-12
feat: more robust parser to get suggested actions

SuZhou-Joe:
### Description
The suggested actions tool may output a response like
```
Here are result { "response": ["{1}", "{2}"], "foo": "bar" }
```
We need a more robust parser to get the json inside the response and make it as our suggestions.

### Issues Resolved

### DIFF
--- a/server/parsers/basic_input_output_parser.ts
+++ b/server/parsers/basic_input_output_parser.ts
@@ -5,7 +5,8 @@
 
 import createDOMPurify from 'dompurify';
 import { JSDOM } from 'jsdom';
-import { IInput, IOutput, Interaction } from '../../common/types/chat_saved_object_attributes';
+import { IInput, IOutput } from '../../common/types/chat_saved_object_attributes';
+import { MessageParser } from '../types';
 
 const sanitize = (content: string) => {
   const window = new JSDOM('').window;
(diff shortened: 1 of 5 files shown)


server/parsers/basic_input_output_parser.ts
```diff
@@ -13,19 +13,79 @@ const sanitize = (content: string) => {
   return DOMPurify.sanitize(content, { FORBID_TAGS: ['img'] }).replace(/!+\[/g, '[');
 };
 
+const isEveryItemString = (array?: unknown) =>
+  !!(Array.isArray(array) && array.every((item) => typeof item === 'string'));
```
ruanyl, line 17: seems don't need `!!` as the result is boolean already

server/parsers/basic_input_output_parser.ts
```diff
@@ -13,19 +13,79 @@ const sanitize = (content: string) => {
...
+
+export const parseSuggestedActions = (string: string): string[] => {
```
ruanyl, line 19: Nit: maybe rename `string: string` -> `value: string`? as it sounds a bit odd that the variable name and type are the same.

server/parsers/basic_input_output_parser.ts
```diff
@@ -13,19 +13,79 @@ const sanitize = (content: string) => {
   return DOMPurify.sanitize(content, { FORBID_TAGS: ['img'] }).replace(/!+\[/g, '[');
 };
 
+const isEveryItemString = (array?: unknown) =>
```
ruanyl, line 16:
```suggestion
const isEveryItemString = (array?: unknown): array is string[] =>
```

Maybe we can call it `isStringArray`?

server/parsers/basic_input_output_parser.ts
```diff
@@ -13,19 +13,79 @@ const sanitize = (content: string) => {
...
+
+  /**
+   * Use the regular expression to find the JSON substring
+   */
+  const match = string.match(jsonPattern);
+
+  const matchedResult = match && match[0];
+
+  if (!matchedResult) {
+    return [];
```
ruanyl, line 61: Maybe add a test case to cover this line? When the output doesn't contain any json-like content.

server/parsers/basic_input_output_parser.ts
```diff
@@ -13,19 +13,79 @@ const sanitize = (content: string) => {
...
+  const suggestedOutputString = string;
+  let suggestedActions: string[] = [];
+  try {
+    suggestedActions = JSON.parse(suggestedOutputString || '[]');
+  } catch (e) {
+    suggestedActions = [];
+  }
+
+  if (suggestedActions.length) {
+    return suggestedActions;
```
wanglam, line 32: Shall we check every item is string string here?

server/parsers/basic_input_output_parser.ts
```diff
@@ -13,19 +13,79 @@ const sanitize = (content: string) => {
...
+   *      (?:          // Start a non-capturing group
+   *        \{         // Match an opening curly brace
+   *        [^{}]*     // Match any characters that are not curly braces (allowing nested structures)
+   *        \}         // Match a closing curly brace
+   *      )
+   *    )*             // Repeat the non-capturing group zero or more times (allowing nested structures)
+   *  \}/g             // Match a closing curly brace, and 'g' flag for global search
+   *
+   */
+  const jsonPattern = /\{(?:[^{}]|(?:\{[^{}]*\}))*\}/g;
```
wanglam, line 51: How about add response key to the jsonPattern? We only want to use this key in the final steps.

SuZhou-Joe: You mean only parse suggested actions in latest interaction right? Good suggestion and will do that.

server/parsers/basic_input_output_parser.ts
SuZhou-Joe: Yeah, we should.

server/parsers/basic_input_output_parser.ts
wanglam: I mean we can match with the response key like `'{"response":'`. It will be more accurate.

SuZhou-Joe: Agree, and add a check to only parse suggested actions for the last interaction.

server/parsers/basic_input_output_parser.ts
SuZhou-Joe: Done

server/parsers/basic_input_output_parser.ts
SuZhou-Joe: Done and thanks for catching.

server/parsers/basic_input_output_parser.ts
SuZhou-Joe: Sure, done.

server/parsers/basic_input_output_parser.ts
SuZhou-Joe: Done.

ruanyl (approved):

wanglam (approved):

joshuali925:
is there some prompt difference that causes agent framework to not output a consistent format for suggestions?

xluo-aws:
Will this be merged to main?

SuZhou-Joe:
It will, there will be a following PR to merge commits from feature/agent-framework to main.

Issue #72, opened 2023-12-13
Pending items for experimental release in 2.12


joshuali925:
This issue tracks the pending items after agent framework implementation is merged to `main` branch in #66 

- [x] merge remaining commits in agent framework using rebase: https://github.com/opensearch-project/dashboards-assistant/pull/71
   - after this, all PRs should change base branch from agent framework to main
- [x] create 2.x branch
- [x] bump version to 3.0.0 to be consistent with other plugins
- [x] add assistant to [3.0.0 build manifest](https://github.com/opensearch-project/opensearch-build/blob/main/manifests/3.0.0/opensearch-dashboards-3.0.0.yml)
   - depends on version bump to 3.0.0 first
- [x] add assistant to opensearch-plugins [.meta](https://github.com/opensearch-project/opensearch-plugins/blob/main/dashboards-plugins/.meta) and [.gitignore](https://github.com/opensearch-project/opensearch-plugins/blob/main/dashboards-plugins/.gitignore)
- [x] follow experimental release guidelines
- [x] fix lint checker https://github.com/opensearch-project/dashboards-assistant/pull/67#discussion_r1424820974
   - PR: https://github.com/opensearch-project/dashboards-assistant/pull/73
- [x] update getting started guide when ready https://github.com/opensearch-project/dashboards-assistant/pull/67#discussion_r1424807379, https://github.com/opensearch-project/dashboards-assistant/pull/123

joshuali925:
@SuZhou-Joe @ruanyl To add the assistant plugin to build manifest, we need to bump plugin version to 3.0.0 in main branch like other plugins. Is it ok or should we wait on this? I'm not sure if ml-commons agent framework is working in 3.0.0

PR #73, opened 2023-12-13
update link checker exclude patterns

joshuali925:
### Description
- exclude urls that end with '$', which are regexes
- remove unnecessary cloning of whole OSD repo

### Issues Resolved
partially #72 

### Check List
  - [ ] New functionality has javadoc added
- [ ] Commits are signed per the DCO using --signoff

### DIFF
--- a/.github/workflows/links_checker.yml
+++ b/.github/workflows/links_checker.yml
@@ -18,7 +18,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: Load Excludes
       run: |
-        LYCHEE_EXCLUDE=$(curl -fsSL https://raw.githubusercontent.com/opensearch-project/OpenSearch-Dashboards/HEAD/.lycheeexclude | cat - .lycheeexclude | grep -v '^#\|^$' | sed -e :a  -e 'N;s/\n/ --exclude /;ta')
+        LYCHEE_EXCLUDE=$(grep -v '^#\|^$' .lycheeexclude | sed -e :a  -e 'N;s/\n/ --exclude /;ta')
         echo "LYCHEE_EXCLUDE=$LYCHEE_EXCLUDE" >> $GITHUB_ENV
     - name: Lychee Link Checker
       id: lychee
--- a/.github/workflows/links_checker.yml
+++ b/.github/workflows/links_checker.yml
@@ -18,14 +18,13 @@ jobs:
     - uses: actions/checkout@v2
     - name: Load Excludes
       run: |
-        LYCHEE_EXCLUDE=$(curl -fsSL https://raw.githubusercontent.com/opensearch-project/OpenSearch-Dashboards/HEAD/.lycheeexclude | grep -v '^#\|^$' | sed -e :a  -e 'N;s/\n/ --exclude /;ta')
+        LYCHEE_EXCLUDE=$(curl -fsSL https://raw.githubusercontent.com/opensearch-project/OpenSearch-Dashboards/HEAD/.lycheeexclude | cat - .lycheeexclude | grep -v '^#\|^$' | sed -e :a  -e 'N;s/\n/ --exclude /;ta')
         echo "LYCHEE_EXCLUDE=$LYCHEE_EXCLUDE" >> $GITHUB_ENV
     - name: Lychee Link Checker
       id: lychee
       uses: lycheeverse/lychee-action@v1.0.9
       with:
-        # exclude urls that end with '$', which are regexes
-        args: --accept=200,403,429 --exclude https://.+\$$ --exclude ${{ env.LYCHEE_EXCLUDE }} --exclude-mail "**/*.html" "**/*.md" "**/*.txt" "**/*.json" "**/*.js" "**/*.ts" "**/*.tsx"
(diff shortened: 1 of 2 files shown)


.github/workflows/links_checker.yml
```diff
@@ -16,22 +16,16 @@ jobs:
...
       run: |
-        LYCHEE_EXCLUDE=$(sed -e :a  -e 'N;s/\n/ --exclude /;ta' OpenSearch-Dashboards/.lycheeexclude)
+        LYCHEE_EXCLUDE=$(curl -fsSL https://raw.githubusercontent.com/opensearch-project/OpenSearch-Dashboards/HEAD/.lycheeexclude | cat - .lycheeexclude | grep -v '^#\|^$' | sed -e :a  -e 'N;s/\n/ --exclude /;ta')
         echo "LYCHEE_EXCLUDE=$LYCHEE_EXCLUDE" >> $GITHUB_ENV
     - name: Lychee Link Checker
       id: lychee
       uses: lycheeverse/lychee-action@v1.0.9
       with:
-        args: --accept=200,403,429 --exclude ${{ env.LYCHEE_EXCLUDE }} --exclude-mail "**/*.html" "**/*.md" "**/*.txt" "**/*.json" "**/*.js" "**/*.ts" "**/*.tsx"
+        args: --accept=200,403,429 --exclude --exclude ${{ env.LYCHEE_EXCLUDE }} --exclude-mail "**/*.html" "**/*.md" "**/*.txt" "**/*.json" "**/*.js" "**/*.ts" "**/*.tsx"
```
ruanyl, line 27: just curious why adding another `--exclude`?

.github/workflows/links_checker.yml
```diff
@@ -16,22 +16,16 @@ jobs:
...
-    - name: Checkout OpenSearch-Dashboards
-      uses: actions/checkout@v2
-      with:
-        repository: opensearch-project/OpenSearch-Dashboards
-        ref: main
-        path: OpenSearch-Dashboards
     - name: Load Excludes
       run: |
-        LYCHEE_EXCLUDE=$(sed -e :a  -e 'N;s/\n/ --exclude /;ta' OpenSearch-Dashboards/.lycheeexclude)
+        LYCHEE_EXCLUDE=$(curl -fsSL https://raw.githubusercontent.com/opensearch-project/OpenSearch-Dashboards/HEAD/.lycheeexclude | cat - .lycheeexclude | grep -v '^#\|^$' | sed -e :a  -e 'N;s/\n/ --exclude /;ta')
```
ruanyl, line 21: Shall we have a repo-specific `.lycheeexclude` instead of rely on the external? I guess we can start with a small list of `.lycheeexclude` first?

.github/workflows/links_checker.yml
joshuali925: accidentally added, removed now. thanks

.github/workflows/links_checker.yml
joshuali925: it's already in the PR
https://github.com/opensearch-project/dashboards-assistant/blob/baa1ab9539e7605a24722ce35071c97a83657a4d/.lycheeexclude#L1-L5

ruanyl: Yes, but do we still need to download from `https://raw.githubusercontent.com/opensearch-project/OpenSearch-Dashboards/HEAD/.lycheeexclude`?

Hailong-am (approved):

.github/workflows/links_checker.yml
Hailong-am: we don't want to maintain common excludes in two places, the mainly part should from OSD core and we have patch in this repo.

ruanyl: > > Yes, but do we still need to download from `https://raw.githubusercontent.com/opensearch-project/OpenSearch-Dashboards/HEAD/.lycheeexclude`?
> 
> we don't want to maintain common excludes in two places, the mainly part should from OSD core and we have patch in this repo.

But the `.lycheeexclude` in OSD repo is not common excludes as I understand, it's specific to OSD repo, am I right?

Hailong-am: > it's specific to OSD repo

agree, it is specific to OSD. But dashboards-assistant is a plugin of OSD, `.lycheeexclude` defined in OSD should suitable for plugins.

joshuali925: i think the OSD list makes sense, but we are the only one using it besides OSD. It was added by https://github.com/opensearch-project/dashboards-assistant/commit/db292f8e115f327db14d1c0d6e4431a30fe0b819, my PR just avoids the whole repo clone for faster CI and debugging. @SuZhou-Joe any specific reason that you added the `.lycheeexclude` from OSD?

SuZhou-Joe: Actually I copy the lint-checker from OSD, and thus I use the .lycheeexclude from OSD as a reference to keep dashboards-assistant align with OSD. It is OK that we remove this step or merge some repo-specific exclude config with OSD's.

SuZhou-Joe (approved):

.github/workflows/links_checker.yml
joshuali925: i see, in that case i'll remove it. the OSD one contains links like `github.com`, it excludes 404 checks for github links in this repo. without it CI would be able to catch outdated github URLs

PR #76, opened 2023-12-15
Add unit test for generic reducer and session hooks

wanglam:
### Description
1. Add unit tests for genericReducer
2. Add unit tests for useDeleteSession and usePatchSession

### Issues Resolved

### Check List
  - [ ] New functionality has javadoc added
- [ ] Commits are signed per the DCO using --signoff

### DIFF
--- a/public/hooks/__tests__/use_sessions.test.ts
+++ b/public/hooks/__tests__/use_sessions.test.ts
@@ -1,0 +1,180 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { useDeleteSession, usePatchSession } from '../use_sessions';
+import { renderHook, act } from '@testing-library/react-hooks';
+import { useCore } from '../../contexts/core_context';
+import { HttpHandler } from '../../../../../src/core/public';
+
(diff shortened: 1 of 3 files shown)


Hailong-am (approved):

public/hooks/__tests__/use_sessions.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    });
+  });
+
+  it('should return data after delete success', async () => {
+    const { result, waitFor } = renderHook(() => useDeleteSession());
+    useCoreMocked.mock.results[0].value.services.http.delete.mockReturnValue(
+      Promise.resolve('deleted')
+    );
+
+    await act(async () => {
```
ruanyl, line 48: Q: curious when we should use `await act`?

wanglam: The `await act` can be removed here. I think we may need `await act` for async callbacks.
If we remove `await act` and `await waitFor` in this test case, the console will output a warning like this:
image

So I think we need add `await` prefix before `act` when ends with non-async actions.

ruanyl (approved):

raintygao (approved):

PR #86, opened 2023-12-27
Search root agent id by rootAgentName specified in opensearch_dashboards.yml

gaobinlong:
### Description
This PR aims to change the behavior of getting root agent id, users have to specify the name of the root agent in the yml, and when they send the first message in the chat box, the node server will call the search agent api in ml-commons to get the root agent id, if multiple root agents are found, we only fetch the latest one. 

This PR also lets users to change the root agent dynamically, they can delete the old root agent firstly and create a new one with the old name,  when sending message, the api in backend will throw 404 error with message `resource_not_found_exception`, this time the node server will retry to search the root agent by the specified name, if found, the new root agent id will be used and cached.

### Check List
  - [x] New functionality has javadoc added
- [x] Commits are signed per the DCO using --signoff

### DIFF
--- a/server/plugin.ts
+++ b/server/plugin.ts
@@ -16,7 +16,7 @@ import { setupRoutes } from './routes/index';
 import { AssistantPluginSetup, AssistantPluginStart, MessageParser } from './types';
 import { BasicInputOutputParser } from './parsers/basic_input_output_parser';
 import { VisualizationCardParser } from './parsers/visualization_card_parser';
-import { AgentIdNotFoundError } from './routes/chat_routes';
+import { AgentNameNotFoundError } from './routes/chat_routes';
 
 export class AssistantPlugin implements Plugin<AssistantPluginSetup, AssistantPluginStart> {
   private readonly logger: Logger;
@@ -37,8 +37,8 @@ export class AssistantPlugin implements Plugin<AssistantPluginSetup, AssistantPl
      * Check if user enable the chat without specifying a root agent id.
      * If so, gives a warning for guidance.
      */
-    if (config.chat.enabled && !config.chat.rootAgentId) {
-      this.logger.warn(AgentIdNotFoundError);
+    if (config.chat.enabled && !config.chat.rootAgentName) {
+      this.logger.warn(AgentNameNotFoundError);
     }
 
     const router = core.http.createRouter();
@@ -53,7 +53,7 @@ export class AssistantPlugin implements Plugin<AssistantPluginSetup, AssistantPl
     // Register server side APIs
     setupRoutes(router, {
       messageParsers: this.messageParsers,
-      rootAgentId: config.chat.rootAgentId,
+      rootAgentName: config.chat.rootAgentName,
     });
 
     core.capabilities.registerProvider(() => ({
--- a/server/plugin.ts
+++ b/server/plugin.ts
@@ -34,7 +34,7 @@ export class AssistantPlugin implements Plugin<AssistantPluginSetup, AssistantPl
       .toPromise();
 
     /**
-     * Check if user enable the chat without specifying a root agent id.
+     * Check if user enable the chat without specifying a root agent name.
      * If so, gives a warning for guidance.
      */
     if (config.chat.enabled && !config.chat.rootAgentName) {
--- a/server/routes/chat_routes.ts
+++ b/server/routes/chat_routes.ts
@@ -37,8 +37,8 @@ const llmRequestRoute = {
 };
 export type LLMRequestSchema = TypeOf<typeof llmRequestRoute.validate.body>;
 
-export const AgentIdNotFoundError =
-  'rootAgentId is required, please specify one in opensearch_dashboards.yml';
+export const AgentNameNotFoundError =
+  'rootAgentName is required, please specify one in opensearch_dashboards.yml';
 
 const getSessionRoute = {
   path: `${ASSISTANT_API.SESSION}/{sessionId}`,
@@ -135,7 +135,8 @@ export function registerChatRoutes(router: IRouter, routeOptions: RoutesOptions)
       context.core.opensearch.client.asCurrentUser,
       routeOptions.messageParsers
     );
-  const createChatService = () => new OllyChatService();
+  const createChatService = (context: RequestHandlerContext) =>
+    new OllyChatService(context, routeOptions.rootAgentName!);
 
   router.post(
     llmRequestRoute,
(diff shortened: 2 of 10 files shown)


server/plugin.ts
```diff
@@ -37,8 +37,8 @@ export class AssistantPlugin implements Plugin<AssistantPluginSetup, AssistantPl
      * Check if user enable the chat without specifying a root agent id.
```
SuZhou-Joe, line 37:
```suggestion
     * Check if user enable the chat without specifying a root agent name.
```

server/services/chat/olly_chat_service.ts
```diff
@@ -114,12 +185,17 @@ export class OllyChatService implements ChatService {
...
   abortAgentExecution(sessionId: string) {
     if (OllyChatService.abortControllers.has(sessionId)) {
       OllyChatService.abortControllers.get(sessionId)?.abort();
     }
   }
+
+  // only used for test
+  public resetRootAgentId() {
+    OllyChatService.rootAgentId = undefined;
+  }
```
SuZhou-Joe, line 200: If only for test, maybe we can initialize a new instance of `OllyChatService` in beforeEach?

gaobinlong: `rootAgentId` is static, the value is used by all instances of `OllyChatService`.

SuZhou-Joe: Oh, I missed that. Then not able to come up with better solution to get rid of this method.

server/routes/chat_routes.ts
```diff
@@ -329,7 +326,7 @@ export function registerChatRoutes(router: IRouter, routeOptions: RoutesOptions)
       request,
       response
     ): Promise<IOpenSearchDashboardsResponse<HttpResponsePayload | ResponseError>> => {
-      const chatService = createChatService();
+      const chatService = createChatService(context);
 
       try {
         chatService.abortAgentExecution(request.body.sessionId);
```
wanglam, line 332: It seems that we can refactor the `abortAgentExecution` to a static method. So we don't need to create ChatService instance anymore.

server/routes/chat_routes.ts
```diff
@@ -135,7 +135,8 @@ export function registerChatRoutes(router: IRouter, routeOptions: RoutesOptions)
       context.core.opensearch.client.asCurrentUser,
       routeOptions.messageParsers
     );
-  const createChatService = () => new OllyChatService();
+  const createChatService = (context: RequestHandlerContext) =>
```
wanglam, line 138: Prefer refactor `createChatService` to 
```ts
 const createChatService = (context: RequestHandlerContext, rootAgentName: string) =>  new OllyChatService(context, routeOptions.rootAgentName)
```
It's dangerous to use `!`.

server/routes/chat_routes.ts
```diff
@@ -144,29 +145,25 @@ export function registerChatRoutes(router: IRouter, routeOptions: RoutesOptions)
...
-        context.assistant_plugin.logger.error(AgentIdNotFoundError);
-        return response.custom({ statusCode: 400, body: AgentIdNotFoundError });
+      if (!routeOptions.rootAgentName) {
+        context.assistant_plugin.logger.error(AgentNameNotFoundError);
+        return response.custom({ statusCode: 400, body: AgentNameNotFoundError });
       }
       const { messages = [], input, sessionId: sessionIdInRequestBody } = request.body;
       const storageService = createStorageService(context);
-      const chatService = createChatService();
+      const chatService = createChatService(context);
```
wanglam, line 154: Since we've already check `routeOptions.rootAgentName` is not empty, we can just pass it here.

server/services/chat/olly_chat_service.ts
```diff
@@ -21,22 +25,89 @@ const INTERACTION_ID_FIELD = 'parent_interaction_id';
...
+      });
+
+      if (!response || response.body.hits.total.value === 0) {
+        throw new Error('cannot find any root agent by name: ' + rootAgentName);
+      }
+      return response.body.hits.hits[0]._id;
+    } catch (error) {
+      let errorMessage = JSON.stringify(error.meta?.body);
+      if (!errorMessage) {
+        errorMessage = error.toString();
```
wanglam, line 66: Since error is any here. The `error` may doesn't have `toString` method. How about refactor the errorMessage like this:
```ts
 const errorMessage = JSON.stringify(error.meta?.body || error);
```

server/services/chat/olly_chat_service.ts
```diff
@@ -21,22 +25,89 @@ const INTERACTION_ID_FIELD = 'parent_interaction_id';
...
-    payload: AgentRunPayload,
-    context: RequestHandlerContext
-  ) {
+  constructor(
+    private readonly context: RequestHandlerContext,
+    private readonly rootAgentName: string
+  ) { }
+
+  private async initRootAgent() {
+    OllyChatService.rootAgentId = await this.searchRootAgent(this.rootAgentName);
```
wanglam, line 36: The `initRootAgent` will be called during the `searchRootAgent` calling the ml-commons API. That will cause unnecessary `searchRootAgent` called. Shall we store the first `searchRootAgent` promise and reuse it?

server/services/chat/olly_chat_service.ts
gaobinlong: Thanks, by my test, if `error` is a instance of `Error`, then JSON.stringify(error) returns `{}` which is not as expected, but I found that `JSON.stringify(error.meta?.body) || error` works well.

server/services/chat/olly_chat_service.ts
gaobinlong: `rootAgentId` is static, it's set when the first call triggers and will not be reset unless encounters a 404 error, we have this check below:
```
if (!OllyChatService.rootAgentId) {
      await this.initRootAgent();
      this.context.assistant_plugin.logger.info(
        `root agent id has not been initialized yet, init it at the first time, current root agent id is:${OllyChatService.rootAgentId}`
      );
    }
```

wanglam: The `OllyChatService.rootAgentId` will still be empty If the first request not response , right? If we call the `requestLLM` or `regenerate` at that time, the  `OllyChatService.rootAgentId` is still empty and will call `initRootAgent` again. This's not a big issue. I'm open to add todo here and fixed it in the future version.

server/routes/chat_routes.ts
gaobinlong: The method `abortAgentExecution ` here implements the method in the interface `ChatService`, it may not be a static method?

server/services/chat/olly_chat_service.ts
gaobinlong: Gotcha, thanks, I'll think about how to implement that.

server/routes/chat_routes.ts
wanglam: It seems the abortAgentExecution won't abort the agent for now. It's ok to me to keep current implementation. Maybe the abort method will depends on context or agentName in the future.

wanglam (approved):

raintygao (approved):

server/services/chat/olly_chat_service.ts
gaobinlong: Will do this in next PR, thanks!

PR #94, opened 2024-01-08
feat: some optimization for customized render function

### LINKED ISSUE #95: [FEATURE] Integrate data grid into Chatbot by using PPL query.
**Do you have any additional context?**
https://github.com/opensearch-project/dashboards-observability/issues/1343

SuZhou-Joe:
### Description
- Export some essential types.
- Container 100% width judgement adjustment.
- Remove ppl visualization related code.
- Add chatContext and props into render context to customized render function.

### Issues Resolved
#95 

### Check List
  - [x] All tests pass, including unit test, integration test and doctest
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added
  - [x] New functionality has user manual doc added
- [x] Commits are signed per the DCO using --signoff

### DIFF
--- a/common/types/chat_saved_object_attributes.ts
+++ b/common/types/chat_saved_object_attributes.ts
@@ -47,7 +47,7 @@ export interface IOutput {
   content: string;
   suggestedActions?: ISuggestedAction[];
   messageId?: string;
-  isVisualization?: boolean;
+  fullWidth?: boolean;
 }
 export type IMessage = IInput | IOutput;
 
--- a/common/types/chat_saved_object_attributes.ts
+++ b/common/types/chat_saved_object_attributes.ts
@@ -43,11 +43,11 @@ export interface IOutput {
   type: 'output';
   traceId?: string; // used for tracing agent calls
   toolsUsed?: string[];
-  // TODO: ppl_visualization type may need to be removed in the PR which replaces ppl query render from visualization to data grid. @suzhou
-  contentType: 'error' | 'markdown' | 'visualization' | 'ppl_visualization' | 'ppl_data_grid';
+  contentType: 'error' | 'markdown' | 'visualization' | 'ppl_data_grid' | string;
   content: string;
   suggestedActions?: ISuggestedAction[];
   messageId?: string;
+  isVisualization?: boolean;
 }
 export type IMessage = IInput | IOutput;
 
--- a/common/types/chat_saved_object_attributes.ts
+++ b/common/types/chat_saved_object_attributes.ts
@@ -43,7 +43,7 @@ export interface IOutput {
   type: 'output';
   traceId?: string; // used for tracing agent calls
   toolsUsed?: string[];
-  contentType: 'error' | 'markdown' | 'visualization' | 'ppl_data_grid' | string;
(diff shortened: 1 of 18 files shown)


common/types/chat_saved_object_attributes.ts
```diff
@@ -43,11 +43,11 @@ export interface IOutput {
   type: 'output';
   traceId?: string; // used for tracing agent calls
   toolsUsed?: string[];
-  // TODO: ppl_visualization type may need to be removed in the PR which replaces ppl query render from visualization to data grid. @suzhou
-  contentType: 'error' | 'markdown' | 'visualization' | 'ppl_visualization' | 'ppl_data_grid';
+  contentType: 'error' | 'markdown' | 'visualization' | 'ppl_data_grid' | string;
```
wanglam, line 46: Shall we use `string` directly?

public/tabs/chat/messages/message_content.tsx
```diff
@@ -37,18 +37,15 @@ export const MessageContent: React.FC<MessageContentProps> = React.memo((props)
...
-        <div className="llm-chat-visualizations">{render({ query: props.message.content })}</div>
-      );
-    }
-
     // content types registered by plugins unknown to assistant
     default: {
       const message = props.message as IMessage;
-      return chatContext.contentRenderers[message.contentType]?.(message.content) ?? null;
+      return (
+        chatContext.contentRenderers[message.contentType]?.(message, {
```
wanglam, line 44: Shall we pass `message.content` here?

common/types/chat_saved_object_attributes.ts
SuZhou-Joe: I'd like to keep some default enum here to indicate these are built-in contentTypes, but in that case maybe we should remove `ppl_data_grid`?

public/tabs/chat/messages/message_content.tsx
SuZhou-Joe: I wandered around what we should pass here for quite a long time and finally decide to pass message here instead of `message.content` because I think `message` contains all the infos a contentType needs to render, and for message_content and customized render function, they should have the same context to render itself, instead of only a field `message.content`. It is a good chance to change what we will pass here as data grid is the first customized contentType.

wanglam: Shall we refactor `contentRenderers` to `messageRenderers`? Or we need to update the argument type at `ContentRenderer` in `public/types.ts`. We may change `content: unknown` to `message: IOutput`.

SuZhou-Joe: Agree, but what about `message: IMessage`?

public/tabs/chat/messages/message_content.test.tsx
```diff
@@ -102,6 +87,6 @@ describe('<MessageContent />', () => {
         }}
       />
     );
-    expect(customizedRenderMock).toHaveBeenCalledWith('mock customized content');
+    expect(customizedRenderMock.mock.calls[0]).toMatchInlineSnapshot();
```
joshuali925, line 90: empty inline snapshot?

wanglam (approved):

PR #104, opened 2024-01-19
fix: suggestedAction could be null

SuZhou-Joe:
### Description
Make suggestedAction more robusted.

### Issues Resolved

### Check List

### DIFF
--- a/server/parsers/basic_input_output_parser.ts
+++ b/server/parsers/basic_input_output_parser.ts
@@ -29,6 +29,14 @@ export const parseSuggestedActions = (value: string): string[] => {
     suggestedActions = [];
   }
 
+  /**
+   * The suggestedOutputString may be 'null'
+   * Fail fast here.
+   */
+  if (!suggestedActions) {
+    return [];
+  }
+
   if (suggestedActions.length) {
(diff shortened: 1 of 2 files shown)


server/parsers/basic_input_output_parser.ts
```diff
@@ -29,6 +29,14 @@ export const parseSuggestedActions = (value: string): string[] => {
...
 
+  /**
+   * The suggestedOutputString may be 'null'
+   * Fail fast here.
+   */
+  if (!suggestedActions) {
+    return [];
+  }
+
   if (suggestedActions.length) {
```
wanglam, line 40: Shall we check if `suggestedActions` was an array here? So we won't need to add `!suggestedActions` above. Just consider if `suggestedOutputString` would be a `number` or `boolean`, access `length` attribute would throw error.

SuZhou-Joe: Actually, if suggestedActions is not an array, it will return [] in line 45.

SuZhou-Joe: I got your point, let me change the block here to single `isStringArray`.

PR #131, opened 2024-02-05
feat: set verbose to false

SuZhou-Joe:
### Description
https://github.com/opensearch-project/ml-commons/issues/1988
Agent framework used to return memoryId and interactionId only when verbose is set to true. But when we set verbose to true, some useless message will be write into additional_info, which is unnecessary and may cause issue like too many payload was passed when updating a doc.

As now agent framework already merge the code to return both when verbose is false, we need to change the flag in Chatbot.

### Issues Resolved

### Check List


wanglam:
@SuZhou-Joe Do we still need the `memoryId` and `interactionId`?  How do we get `memoryId` and `interactionId` after set `verbose` to `false`?

SuZhou-Joe:
@wanglam We need the memoryId and interactionId, ml-commons has merged a PR to return memory id and interaction id when verbose is false.

wanglam (approved):

ruanyl (approved):

PR #159, opened 2024-02-26
Add workflow to verify binary install into dashboards

derek-ho:
### Description
I have observed some issues within other plugins of issues being only caught at run time. This is because several things can go wrong during the build process, which may not be caught in a dev setup. This adds a workflow to verify that building and installing into OSD works on every PR.

Related issues:
https://github.com/opensearch-project/security-dashboards-plugin/issues/1709
https://github.com/opensearch-project/security-analytics-dashboards-plugin/pull/875
https://github.com/opensearch-project/dashboards-observability/pull/309
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5952

### Issues Resolved

### Check List

### DIFF
--- a/.github/workflows/verify-binary-installation.yml
+++ b/.github/workflows/verify-binary-installation.yml
@@ -1,0 +1,55 @@
+name: 'Install Dashboards with Plugin via Binary'
+
+on: [push, pull_request]
+env:
+  OPENSEARCH_VERSION: '3.0.0'
+  CI: 1
+  # avoid warnings like "tput: No value for $TERM and no -T specified"
+  TERM: xterm
+
+jobs:
+  verify-binary-installation:
+    name: Run binary installation
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


derek-ho:
Testing to see if this would have caught RC testing issue by reverting PR fix in: [6a480cd](https://github.com/opensearch-project/dashboards-assistant/pull/159/commits/6a480cd4e40618fa932efc669b5ee8f3d81793e4)

derek-ho:
Actually not sure if this would have caught it since reading more about the issue it may be a platform specific thing that caused a different version of node to be used. @joshuali925 can you take a look at this PR regardless? I will revert the revert commit if you are in agreement that this wouldn't have caused it (although we can test it by re-running the job after OSD core no longer has issues)

joshuali925 (approved):
thanks. i heard those were caused by newer nodejs versions running on older platforms, i don't think `ubuntu-latest` has the issue. but i think this is good to have

raintygao:
@derek-ho Should this change be released in 2.13?

raintygao (approved):

Issue #156, opened 2024-02-22
[FEATURE] Support and integrate sidecar for chatbot container


raintygao:
**Is your feature request related to a problem?**
We want to update the container of chatbot into a sidecar service.

**What solution would you like?**
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5620 We will create a new PR to OSD to support sidecar service firstly and integrate it in chatbot once PR merged.
The sidecar has 3 variations: Docked left, right, and takeover.
![image](https://github.com/opensearch-project/dashboards-assistant/assets/42465692/4f8f5d04-496a-47bf-b839-b77ee14a4af5)
![image](https://github.com/opensearch-project/dashboards-assistant/assets/42465692/bf603a80-e3cb-4f55-8c91-2b81516edf68)
![image](https://github.com/opensearch-project/dashboards-assistant/assets/42465692/fe6b6676-411d-43a6-94f7-0309ba334d7f)

PR #164, opened 2024-03-13
Integrate sidecar service

raintygao:
### Description
Integrate sidecar service in OSD

### Issues Resolved
https://github.com/opensearch-project/dashboards-assistant/issues/156
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5620
Depending on https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5920

### Check List


raintygao:
Depending on https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5920, so CI would fail for now.

public/chat_header_button.tsx
```diff
@@ -119,12 +122,36 @@ export const HeaderChatButton = (props: HeaderChatButtonProps) => {
...
+          },
+        });
+        flyoutLoaded = true;
+      }
+    } else if (flyoutLoaded && flyoutVisible) {
+      core.overlays.sidecar().show();
+    } else if (flyoutLoaded && !flyoutVisible) {
+      core.overlays.sidecar().hide();
+    }
+  }, [flyoutVisible, flyoutLoaded, flyoutMountPoint]);
```
wanglam, line 143: `flyoutMountPoint` is a ref, we don't need to add it to the dependencies array.

public/chat_header_button.tsx
```diff
@@ -119,12 +122,36 @@ export const HeaderChatButton = (props: HeaderChatButtonProps) => {
...
 
+  useEffect(() => {
+    if (!flyoutLoaded && flyoutVisible) {
+      const mountPoint = flyoutMountPoint.current;
+      if (mountPoint) {
+        core.overlays.sidecar().open(mountPoint, {
+          className: 'chatbot-sidecar',
+          config: {
+            dockedMode: SIDECAR_DOCKED_MODE.RIGHT,
+            paddingSize: 460,
```
wanglam, line 133: Shall we store 460 in some place? Just find it appear many times in this PR.

public/chat_header_button.tsx
```diff
@@ -119,12 +122,36 @@ export const HeaderChatButton = (props: HeaderChatButtonProps) => {
...
+  }, [flyoutVisible, flyoutLoaded, flyoutMountPoint]);
+
   const onKeyUp = (e: React.KeyboardEvent<HTMLInputElement>) => {
     if (e.key === 'Escape') {
       inputRef.current?.blur();
     }
   };
 
+  const setMountPoint = useCallback((mountPoint) => {
+    flyoutMountPoint.current = mountPoint;
```
wanglam, line 152: Shall we check if flyoutVisible here. We may need to call `core.overlays.sidecar().open` here if, flyoutVisible is `true` and previous `flyoutMountPoint.current` is null.

public/chat_header_button.test.tsx
```diff
@@ -57,6 +48,30 @@ jest.mock('./services', () => {
...
+          render(<MountWrapper mount={mountPoint} />, {
+            container: attachElement,
+          });
+        },
+        hide: () => {
+          attachElement.style.display = 'none';
+        },
+        show: () => {
+          attachElement.style.display = 'block';
+        },
```
wanglam, line 69: Shall we add some test cases for show and hide function?

public/chat_header_button.tsx
raintygao: Can you elaborate more? I think we don't need to do more judgement in setMountPoint. Some necessary judgements are done when calling `sidecar().open()`

wanglam: The `sidebar().open()` in `useEffect` will be called if `flyoutVisible` equal true and `mountPoint` is truly. It won't be called if the mountPoint is null. Shall we add related logic for this case? The logic should be move to `setMountPoint` if we need to handle this.

public/utils/constants.ts
```diff
@@ -10,3 +11,8 @@ export enum TAB_ID {
   HISTORY = 'history',
   TRACE = 'trace',
 }
+
+export const DEFAULT_SIDECAR_DOCKED_MODE = SIDECAR_DOCKED_MODE.RIGHT;
+export const DEFAULT_SIDECAR_LEFT_OR_RIGHT_SIZE = 460;
+// this is a default padding top size for sidecar when switching to takeover
+export const DEFAULT_SIDECAR_TAKEOVER_PADDING_TOP_SIZE = 460;
```
wanglam, line 18: As I remember, this value should be 136 before. Is this value the same as `DEFAULT_SIDECAR_LEFT_OR_RIGHT_SIZE`?

public/chat_header_button.tsx
```diff
@@ -119,12 +122,36 @@ export const HeaderChatButton = (props: HeaderChatButtonProps) => {
     }
   };
 
+  useEffect(() => {
```
wanglam, line 129: Shall we move `useEffect` to the bottom of `useCallback`?

raintygao: Do you mean `setMountPoint`? I dont think declaration order will have any difference because `SetMountPoint` will always excute after `sidecar().open()`.

wanglam (approved):
LGTM

Issue #154, opened 2024-02-22
[FEATURE]Fetch chatbot root agent id before execution


xluo-aws:
**Is your feature request related to a problem?**
The current implementation is only retrieve root agent id if execution fails with resource node found error. The reason is that the get root agent id API is heavy because it need to loop through all agents to match the root agent name. It impacts user experience, admin need to remember to deprovision the current root agent to make the new one take effect. 

**What solution would you like?**
The new solution is to call get root agent id every time before execute the agent. The get root agent id is enhanced with a very light implementation so performance and latency is no longer a concern.

**What alternatives have you considered?**
Keep the current logic

**Do you have any additional context?**
No

gaobinlong:
Closed by https://github.com/opensearch-project/dashboards-assistant/pull/165.

PR #165, opened 2024-03-13
Fetch root agent id before executing the agent

gaobinlong:
### Description
Fetch root agent id before executing the agent every time, and remove some unused unit test code.

### Issues Resolved
https://github.com/opensearch-project/dashboards-assistant/issues/154

### Check List


lezzago:
Why does this change need to happen? Was this to deal with scenarios when the agent id changes? If so, can we try to refetch the root agent if the call fails to the execute agent api with the old root agent id?

server/services/chat/olly_chat_service.ts
```diff
@@ -58,40 +48,22 @@ export class OllyChatService implements ChatService {
...
-        const oldRootAgentId = OllyChatService.rootAgentId;
-        await this.initRootAgent();
-        this.context.assistant_plugin.logger.info(
-          `cannot find the root agent id: ${oldRootAgentId}, try to fetch the new root agent id, now it is:${OllyChatService.rootAgentId}`
-        );
-        return await this.callExecuteAgentAPI(payload);
-      } else {
-        throw error;
-      }
+      throw error;
```
joshuali925, line 56: since it's just `throw error`, maybe remove the try catch?

joshuali925:
> Why does this change need to happen? Was this to deal with scenarios when the agent id changes? If so, can we try to refetch the root agent if the call fails to the execute agent api with the old root agent id?

@lezzago it's explained in the linked issue, if we cache the agent id, user won't be able to change agent unless they delete the old one. cache was added to save API calls, but the get config API has lower overhead

lezzago:
> But i still think it's a rare operation to change agent ID. although it's not the bottleneck and get call is lightweight, it is still some overhead to every user request

Ah ok, that makes and I agree, the get config api is a lightweight call. We may need to revisit this if we start to see all these overhead calls adds problems, especially if the dashboard assistant apis are getting by outside services and not being used just on OpenSearch Dashboards.

xluo-aws:
> > But i still think it's a rare operation to change agent ID. although it's not the bottleneck and get call is lightweight, it is still some overhead to every user request
> 
> Ah ok, that makes and I agree, the get config api is a lightweight call. We may need to revisit this if we start to see all these overhead calls adds problems, especially if the dashboard assistant apis are getting by outside services and not being used just on OpenSearch Dashboards.

Yes, we can start with this approach. One future optimization is to add cache but it introduces complexity so we don't want to add it until we see this API adds more and more overhead.

server/services/chat/olly_chat_service.ts
gaobinlong: Yeah, I've changed that, thanks!

lezzago (approved):

PR #191, opened 2024-05-27
Add data source service

wanglam:
### Description
This PR is a part of multi data source support, add a new data source service to handle all the data source update logic.
This data source service will implement below functions:
1.Provide a setDataSourceId in setup and start return values. So other plugins can change data source id by this function.
2.Provide a init method, so chat bot can init data source id when dialog open. This init data source id function will check if multi data source enabled. It will try to get selection data source id from dsm or default data source from ui settings.
3.Provide a clearDataSourceId method, it reset data source id to null. It will be called when chatbot dialog been closed.
4.Provide a subscribeDataSourceId method, so chatbot can observe this subject to interact with different data source.
5.Provide a getDataSourceQuery method, so chatbot can send API request with this data source id query object. It will handle the local cluster automaticlly. 

### Issues Resolved
A part of #192

### Check List


public/plugin.tsx
```diff
@@ -131,6 +140,7 @@ export class AssistantPlugin
     }
 
     return {
+      ...dataSourceSetupResult,
```
SuZhou-Joe, line 143:
```suggestion
      dataSourceService: dataSourceSetupResult
```
Is there any place that will use `dataSourceSetupResult`, and if so, can we make it a single entry to keep the export interface clean?

public/plugin.tsx
```diff
@@ -160,8 +170,12 @@ export class AssistantPlugin
     setChrome(core.chrome);
     setNotifications(core.notifications);
 
-    return {};
+    return {
+      ...this.dataSourceService.start(),
```
SuZhou-Joe, line 174:
```suggestion
      dataSourceServiceStart: this.dataSourceService.start(),
```
Same here, I'd vote for a single clear entry.

public/services/data_source_service.ts
```diff
@@ -0,0 +1,128 @@
...
+import type { DataSourceOption } from '../../../../src/plugins/data_source_management/public/components/data_source_menu/types';
+import type { DataSourceManagementPluginSetup } from '../types';
+
+export enum DataSourceIdFrom {
+  UiSettings,
+  DataSourceSelection,
+  Customized,
+}
+
+const getDSMDataSourceSelectionOption = (dataSourceSelection: Map<string, DataSourceOption[]>) => {
```
SuZhou-Joe, line 17: The function name seems not comply with what it is doing. What about `getSingleSelectedDataSourceOption`?

public/plugin.tsx
wanglam: How about just export a `setDataSourceId` method? For now, this method not been used in any place.

public/services/data_source_service.ts
```diff
@@ -0,0 +1,128 @@
...
+          selectedDataSourceOption?.id ?? null,
+          DataSourceIdFrom.DataSourceSelection
+        );
+      });
+
+    this.uiSettings.get$('defaultDataSource', null).subscribe((newDataSourceId) => {
+      if (this.dataSourceIdFrom === DataSourceIdFrom.UiSettings) {
+        this.setDataSourceId(newDataSourceId, DataSourceIdFrom.UiSettings);
+      }
+    });
```
SuZhou-Joe, line 108: Do we need to be aware of the `defaultDataSource`? I thought it should be transparent to plugins.

What about we initialize a hidden data source selector if none picker is mounted to the page? And that picker will help to tell us what the defaultDataSource is.

wanglam: This's a good idea. But there are three limitations after change to a transparent data source picker:
1. onSelectedDataSource in data source picker need to wait all data source fetched, it will take more time to determine the default data source.
2. The `onSelectedDataSource` won't be fired after default data source, the data source id in data source service will still use the old one.
3. Since we are depend on data source selection, the transparent data source picker will add selection value to the data source selection. We need to add some logic to filter out data source selection from this component. @raintygao may need to modify the OSD PR.
I can change to use a transparent data source picker if 2 & 3 have been addressed or skipped. But I think it would be better if we can add some defaultDataSource functions to subscribe the default data source changes in DSM plugin.

raintygao: DSM hasn't provided a way to get default data source. As it's stored in ui settings and DSM also gets this value from uisettings, so i think we don't need to depend on DSM to get this additionally. Im open to be aware of `defaultDataSource` in external plugin in this stage.

public/services/data_source_service.ts
```diff
@@ -0,0 +1,130 @@
...
+        const selectedDataSourceOption = getSingleSelectedDataSourceOption(dataSourceSelection);
+        this.setDataSourceId(
+          selectedDataSourceOption?.id ?? null,
+          DataSourceIdFrom.DataSourceSelection
+        );
+      });
+
+    this.uiSettings.get$('defaultDataSource', null).subscribe((newDataSourceId) => {
+      if (this.dataSourceIdFrom === DataSourceIdFrom.UiSettings) {
+        this.setDataSourceId(newDataSourceId, DataSourceIdFrom.UiSettings);
```
ruanyl, line 108: Q: why do we need to subscribe to the change of `defaultDataSource`? My understanding is `this.setDataSourceId` should be setting the current selected data source, it should not take `defaultDataSource` into consideration.

I guess what we are trying to implement here is when `this.dataSourceId$` is empty, assistant should use `defaultDataSource`, maybe we should incorporate this logic in `getDataSourceId` and `subscribeDataSourceId`?

public/services/data_source_service.ts
```diff
@@ -0,0 +1,130 @@
...
+export class DataSourceService {
+  private dataSourceId$ = new BehaviorSubject<string | null>(null);
+  private uiSettings: IUiSettingsClient | undefined;
+  private dataSourceManagement: DataSourceManagementPluginSetup | undefined;
+  private dataSourceSelectionSubscription: Subscription | undefined;
+  private dataSourceIdFrom: DataSourceIdFrom | undefined;
+
+  constructor() {}
+
+  initDefaultDataSourceIdIfNeed() {
```
ruanyl, line 37: When will this function to be called?

ruanyl: > so chat bot can init data source id when dialog open.

Found in the PR description, but do chatbot really need to call `initDefaultDataSourceIdIfNeed` every time it's opened? I feel it more intuitive if simply `subscribeDataSourceId` or `getDataSourceId` can do it.

wanglam: The motivation of init everytime when chatbot open is get latest default data source id. Since the default data source id may changed in another tabs, add the init code to chatbot open to avoid using the expired default data source id.

public/plugin.tsx
SuZhou-Joe: Agree that.

ruanyl (commented):
I'm trying to describe what I'm thinking via suggested changes, maybe it's more straight-forward to just look at the code.

The key points are:
1. use `dataSourceId$` to store computed value from data source selector
2. derive the data source to be used in chatbot via `getDataSourceId$()`
And chatbot can simply subscribe to `getDataSourceId$()` to get which data source to use. Does it make sense?

ruanyl (approved):

public/services/data_source_service.ts
wanglam: Sure. @raintygao already raised a [PR](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6843) to add these methods in the DSM plugin. Let's keep current uiSettings implementation. Will raised another PR to fix this default data source after @raintygao 's PR merged.

raintygao (approved):

wanglam:
@ruanyl @raintygao I've update the `dataSourceSelection` to optional in `DataSourceManagementPluginSetup`. It's a minor change. Feel free to help me take a look. Thank you.

PR #194, opened 2024-05-30
reset chat or reload history after data source change

### LINKED ISSUE #192: [FEATURE] Support multi data source in assistant
**Is your feature request related to a problem?**
 We will support multi data source in assistant, assistant needs to get the data source selected by other applications from the OSD page and use this data source to connect to the corresponding OS cluster. When the data source selected on the page changes, the assistant should refresh and update accordingly and connect to the new OS cluster.

**What solution would you like?**
image

There are two parts that need to be updated. 

In OSD, since the data source selector and menu components provided by DSM don't store the selected value after the data source is selected, we will add a service in DSM. After the component is selected, store the selected value after selecting data source and delete the selected value after the component is unmounted. Other plugins such as assistant can get the value through this instance.

In assistant, we need to integrate the above new service in DSM, and get selected data source through the data source selector in the OSD page on the front end. Meanwhile, the API needs to be upgraded to support passed data source ID and be able to connect to the corresponding OS cluster through specific data source id.

wanglam:
### Description
1. Reset chat after data source change
2. Reload history and reset search params after data source change
3. Reset to chat tab when in trace tab

### Issues Resolved
#192 

### Check List


public/tabs/history/chat_history_page.tsx
```diff
@@ -96,14 +100,29 @@ export const ChatHistoryPage: React.FC<ChatHistoryPageProps> = React.memo((props
...
+    services.conversations.reload();
+    return () => {
+      services.conversations.abortController?.abort();
+    };
   }, [props.shouldRefresh, services.conversations]);
 
   useEffect(() => {
     services.conversations.load(bulkGetOptions);
+    const subscription = services.dataSource
+      .getDataSourceId$()
```
ruanyl, line 116: Would it be possible that `dataSourceId$` emit the same value one after another? In such case, I think we don't want to reload.

public/tabs/history/chat_history_page.tsx
```diff
@@ -96,14 +100,29 @@ export const ChatHistoryPage: React.FC<ChatHistoryPageProps> = React.memo((props
...
+      .pipe(skip(1))
```
ruanyl, line 117: Could you leave a comment to describe why it should skip the first value?

public/hooks/use_chat_actions.tsx
```diff
@@ -106,6 +106,15 @@ export const useChatActions = (): AssistantActions => {
     }
   };
 
+  const resetChat = () => {
```
ruanyl, line 109: Can you simply call `loadChat` without any parameter instead of creating a new method? we can call `loadChat()` to start a new conversation

wanglam: The `loadChat` only  work when current tab is chat tab. The loadChat will set current tab to CHAT in [this line](https://github.com/opensearch-project/dashboards-assistant/blob/main/public/hooks/use_chat_actions.tsx#L90). I think we can't change tab to CHAT if user is in history tab. So add a new `resetChat` here.

public/tabs/history/chat_history_page.tsx
wanglam: Sure, since the the conversations load will be called after mount. We skip the first value here to avoid duplicate load API request.

public/tabs/history/chat_history_page.tsx
wanglam: In current implementation, the same value will be emit after default data source change. Do you have any suggestions about that? How about add a `distinctUntilChanged`?

public/chat_header_button.tsx
```diff
@@ -54,6 +56,9 @@ export const HeaderChatButton = (props: HeaderChatButtonProps) => {
   const flyoutFullScreen = sidecarDockedMode === SIDECAR_DOCKED_MODE.TAKEOVER;
   const flyoutMountPoint = useRef(null);
 
+  const resetChatRef = useRef(props.assistantActions.resetChat);
+  resetChatRef.current = props.assistantActions.resetChat;
```
SuZhou-Joe, line 60: Can we just use the `props.assistantActions.resetChat`?

public/tabs/history/chat_history_page.tsx
SuZhou-Joe: Can we say that we are skipping the case that dataSource$ emits the same value? In that case we'd use `distinctUntilChanged` instead of skip the first value right?

public/chat_header_button.tsx
wanglam: I've tested in my local machine, the `props.assistantActions.resetChat` will be changed after chat button render. The `getDataSourceId$` will be executed multi times. It's OK to change to use `props.assistantActions.resetChat` directly.

public/tabs/history/chat_history_page.tsx
wanglam: Not the same value. Since `dataSourceId$` and `defaultDataSourceId$` are `BehaviorSubject`.  The observer will be executed once first subscribed. The observer will reload conversations history. Since we already call `conversations.load` after mount, skip first value here to avoid duplicate API call.

raintygao (dismissed):

public/services/data_source_service.ts
```diff
@@ -74,24 +74,23 @@ export class DataSourceService {
   }
 
   setDataSourceId(newDataSourceId: string | null) {
-    if (this.dataSourceId$.getValue() === newDataSourceId) {
-      return;
-    }
```
raintygao, line 79: Could you elaborate more on why we remove duplicate check in set function?

wanglam: Since we already add `distinctUntilChanged` in `getDataSourceId$`, the next value won't be the same. Then we can remove the duplicate check here.

Hailong-am:
how about `how was it generated`(trace) page? I guess that page would need to reset if user in that page and datasource changed.

wanglam:
> how about `how was it generated`(trace) page? I guess that page would need to reset if user in that page and datasource changed.

Do you mean refresh how was it generated page when data source changed? I think it's a little bit confused. Because user won't get the trace info in another data source.

Hailong-am:
> > how about `how was it generated`(trace) page? I guess that page would need to reset if user in that page and datasource changed.
> 
> Do you mean refresh how was it generated page when data source changed? I think it's a little bit confused. Because user won't get the trace info in another data source.

right, show existing trace data maybe confusing. how about reset to chat window?

wanglam:
> > > how about `how was it generated`(trace) page? I guess that page would need to reset if user in that page and datasource changed.
> > 
> > 
> > Do you mean refresh how was it generated page when data source changed? I think it's a little bit confused. Because user won't get the trace info in another data source.
> 
> right, show existing trace data maybe confusing. how about reset to chat window?

Thanks for suggestions, I've already updated this part.

ruanyl (approved):

Hailong-am (approved):

raintygao (approved):

public/tabs/history/chat_history_page.tsx
```diff
@@ -56,11 +57,15 @@ export const ChatHistoryPage: React.FC<ChatHistoryPageProps> = React.memo((props
     }),
     [pageIndex, pageSize, debouncedSearchName]
   );
+  const bulkGetOptionsRef = useRef(bulkGetOptions);
+  bulkGetOptionsRef.current = bulkGetOptions;
```
SuZhou-Joe, line 61: Not needed anymore I think?

wanglam: Sure, can be removed.

public/tabs/history/chat_history_page.tsx
```diff
@@ -96,17 +101,44 @@ export const ChatHistoryPage: React.FC<ChatHistoryPageProps> = React.memo((props
...
+      }
+      if (bulkGetOptions.page !== 1) {
+        setPageIndex(0);
+      }
+    });
     return () => {
       services.conversations.abortController?.abort();
+      subscription.unsubscribe();
     };
   }, [services.conversations, bulkGetOptions]);
```
SuZhou-Joe, line 131:
```suggestion
  const dataSourceUpdate = useObservable(services.dataSource.dataSourceIdUpdates$);

  useEffect(() => {
    setSearchName('');
    setDebouncedSearchName('');
    setPageIndex(0);
  }, [dataSourceUpdate]);

  useEffect(() => {
    services.conversations.load(bulkGetOptions);
    return () => {
      services.conversations.abortController?.abort();
    };
  }, [services.conversations, bulkGetOptions]);
```
What about changing to like this?

SuZhou-Joe (approved):

PR #201, opened 2024-06-04
Hide notebook feature when MDS enabled and remove security dashboard API dependency calling

raintygao:
### Description
Hide notebook feature when MDS enabled and remove security dashboard API dependency calling

### Check List


public/types.ts
```diff
@@ -46,10 +45,6 @@ export interface AssistantSetup {
    * Returns true if chat UI is enabled.
    */
   chatEnabled: () => boolean;
-  /**
-   * Returns true if current user has permission to use assistant features.
-   */
-  userHasAccess: () => Promise<boolean>;
```
SuZhou-Joe, line 52: Is it safe to remove this API?

SuZhou-Joe (approved):

public/services/data_source_service.mock.ts
```diff
@@ -4,12 +4,18 @@
  */
 
 export class DataSourceServiceMock {
-  constructor() {}
+  private _isMDSEnabled = false;
+  constructor(isMDSEnabled?: boolean) {
+    this._isMDSEnabled = isMDSEnabled ?? false;
+  }
 
   getDataSourceQuery() {
```
wanglam, line 12: Shall we resolve an empty object for MDS not enabled?

raintygao: I don't think so. getDataSourceQuery is a function that would be called all time, although it would return empty string when MDS disabled.

server/plugin.ts
```diff
@@ -42,8 +42,9 @@ export class AssistantPlugin implements Plugin<AssistantPluginSetup, AssistantPl
     });
 
     // Register server side APIs
-    setupRoutes(router, {
+    registerChatRoutes(router, {
```
wanglam, line 45: Is it necessary to call `registerChatRoutes` instead of `setupRoutes`?

public/services/data_source_service.mock.ts
wanglam: Just to keep consistent mock logic with the real world in case of some place set `isMDSEnabled` false but the `getDataSourceQuery` still return dataSourceId.

raintygao: I have updated getDataSourceQuery function in mock to make it consistent.

wanglam (approved):

PR #203, opened 2024-06-06
Add patch style for oui fixed components

raintygao:
### Description
There are some UI components from library whose position is fixed and are not compatible with each other and also not compatible with sidecar container. Meanwhile, there is currently no way to provide config for these components at runtime.
This PR adds a hook to patch a style for all these already known components to make them compatible with sidecar.
Will consider to support config provider in OUI and refer this PR later.

### ScreenShot

image
image
image

Before: the left or right property of flyout is always to be 0. The content width of bottomBar is  always be 100%.

image
image
image

This still needs a PR in OSD to update sidecar z-index. https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6964
After: the left or right property of flyout will be updated accordingly. The bottomBar will be added paddingRight or paddingLeft style accordingly.

### Check List


public/hooks/use_patch_fixed_style.ts
```diff
@@ -0,0 +1,89 @@
...
+    style.appendChild(text);
+
+    textRef.current = text;
+
+    const subscription = sidecarConfig$.subscribe((config) => {
+      if (!config) return;
+      updateHeadStyle(config, textRef.current);
+    });
+    return () => {
+      subscription.unsubscribe();
```
wanglam, line 33: Shall we remove appended style after unmount?

ruanyl:
@raintygao Can you also add a screenshot in the description to better show what is the problem?

raintygao:
> @raintygao Can you also add a screenshot in the description to better show what is the problem?

Sure. Updated.

public/chat_header_button.tsx
```diff
@@ -52,6 +53,7 @@ export const HeaderChatButton = (props: HeaderChatButtonProps) => {
   const core = useCore();
   const flyoutFullScreen = sidecarDockedMode === SIDECAR_DOCKED_MODE.TAKEOVER;
   const flyoutMountPoint = useRef(null);
+  usePatchFixedStyle();
```
SuZhou-Joe, line 56: I am thinking, the patched style should only apply when the chatbot is visible right?

raintygao:
A PR in OSD that is related.cc @ruanyl @SuZhou-Joe https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6964

public/chat_header_button.tsx
raintygao: Yes.

raintygao: Patchstyle contains the logic that when chatbot is closed or docked takeover. When in these cases, we will update a empty style to remove patch style.

SuZhou-Joe: I see, so the `sidecarConfig$` will emit a null value once the Chatbot is not visible?

wanglam (approved):

public/chat_header_button.tsx
raintygao: Sidecar config will emit a new value with isHidden property is true.Then we will update inserted style with ""

SuZhou-Joe: I did not find the code that we will update the inserted style when config.hidden is true.

public/hooks/use_patch_fixed_style.ts
```diff
@@ -0,0 +1,88 @@
...
+          padding-left: ${paddingSize}px
+      }
+        `;
+    }
+  }
+
+  // If sidecar closes or docked direction change to takeover mode, we will keep css empty to remove patch style and update.
+  requestAnimationFrame(() => {
+    text.textContent = css;
+  });
```
SuZhou-Joe, line 87:
```suggestion
  requestAnimationFrame(() => {
    if (config?.isHidden) { text.textContent = ''; return ; }
    text.textContent = css;
  });
```

Nit: I'd suggest explicitly making the textContent as empty.

SuZhou-Joe (approved):

public/hooks/use_patch_fixed_style.ts
raintygao: I think here takeover dockedMode also takes the same effect as isHidden, should we also explicitly append?

Issue #216, opened 2024-07-16
[FEATURE] Support creating alert in the chat window


qianheng-aws:
**Is your feature request related to a problem?**
Support creating alert in the chat window as a follow-up question of alert analysis, as shown in this mock up [figma workflow](https://www.figma.com/proto/Y2exyzJHXf5LKITzCNncSL/Olly-Playground-Launch?node-id=5514-148629&t=y613jPVTUyRJMSAK-0&scaling=min-zoom&content-scaling=fixed&page-id=5495%3A25201&starting-point-node-id=5508%3A146289&show-proto-sidebar=1)

**What solution would you like?**

**What alternatives have you considered?**

**Do you have any additional context?**

dblock:
[Catch All Triage - [1](https://github.com/dblock), [2](https://github.com/krisfreedain), [3](https://github.com/amistrn)]

PR #217, opened 2024-07-17
Support creating alert in the chat window

qianheng-aws:
### Description
Support creating alert in the chat window as a follow-up question of alert analysis, as shown in this mock up [figma workflow](https://www.figma.com/proto/Y2exyzJHXf5LKITzCNncSL/Olly-Playground-Launch?node-id=5514-148629&t=y613jPVTUyRJMSAK-0&scaling=min-zoom&content-scaling=fixed&page-id=5495%3A25201&starting-point-node-id=5508%3A146289&show-proto-sidebar=1)

### Issues Resolved
https://github.com/opensearch-project/dashboards-assistant/issues/216

### Check List

### DIFF
--- a/public/chat_flyout.tsx
+++ b/public/chat_flyout.tsx
@@ -12,6 +12,7 @@ import { ChatWindowHeader } from './tabs/chat_window_header';
 import { ChatHistoryPage } from './tabs/history/chat_history_page';
 import { AgentFrameworkTracesFlyoutBody } from './components/agent_framework_traces_flyout_body';
 import { TAB_ID } from './utils/constants';
+import { ChatOverrideHeader } from './tabs/chat_override_header';
 
 interface ChatFlyoutProps {
   flyoutVisible: boolean;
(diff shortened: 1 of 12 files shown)


public/chat_flyout.tsx
```diff
@@ -86,52 +87,55 @@ export const ChatFlyout = (props: ChatFlyoutProps) => {
     >
       <>
         <div className={cs('llm-chat-flyout-header')}>
-          <ChatWindowHeader />
+          {props.overrideComponent ? <ChatOverrideHeader /> : <ChatWindowHeader />}
```
wanglam, line 90: Is this `ChatOverrideHeader` suitable for all override component header? Shall we just hide the `ChatWindowHeader` and let `overrideComponent` take over the whole screen?

server/parsers/create_monitor_parser.ts
wanglam, line 1: Seems this file not been used. Shall we remove it?

public/chat_flyout.tsx
qianheng-aws: Yes, I think this is a standard header for all override component. As UX required, we should change the title, add a back button and hide the history button. There is a mock up for AD, which also applies to alert.

image

PR #268, opened 2024-09-04
Expose a general function for agent execution

ruanyl:
### Description
This PR adds an assistant client to both `public(UI)` and `server` which currently exposes the agent execute API

### Issues Resolved

### Check List

### DIFF
--- a/public/services/assistant_client.ts
+++ b/public/services/assistant_client.ts
@@ -1,0 +1,20 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { API_BASE } from '../../common/constants/llm';
+import { HttpSetup } from '../../../../src/core/public';
+
+export class AssistantClient {
+  constructor(private http: HttpSetup) {}
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  executeAgent = (agentName: string, parameters: Record<string, any>) => {
+    return this.http.fetch({
+      method: 'POST',
+      path: `${API_BASE}/agents/${agentName}/_execute`,
+      body: JSON.stringify(parameters),
+    });
+  };
+}
--- a/public/services/assistant_client.ts
+++ b/public/services/assistant_client.ts
@@ -6,15 +6,20 @@
 import { API_BASE } from '../../common/constants/llm';
 import { HttpSetup } from '../../../../src/core/public';
 
+interface Options {
+  dataSourceId?: string;
+}
+
 export class AssistantClient {
   constructor(private http: HttpSetup) {}
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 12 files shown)


server/services/assistant_client.ts
```diff
@@ -0,0 +1,63 @@
...
+          getClient: (dataSourceId: string) => Promise<OpenSearchClient>;
+        };
+      };
+    }
+  ) {}
+
+  executeAgent = async (
+    agentName: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    parameters: Record<string, any>
```
Hailong-am, line 45: do we want to support executeAgent with agent id?

Hailong-am (approved):

server/services/assistant_client.ts
ruanyl: I also considered this, but we don't have use case at the moment which requires to execute agent with id directly, so I decided to not add it for now, we can add it when it's needed.

qianheng-aws: +1

qianheng-aws: Knowledge base needs to execute by agent id directly

Hailong-am: @qianheng-aws may have a requirement to execute agent directly

chishui (commented):
How do we allow other plugins to overwrite assistantClient and provide their own customized functionalities?

server/services/assistant_client.ts
ruanyl: @Hailong-am @qianheng-aws Updated, could you take another look?

ruanyl:
> How do we allow other plugins to overwrite assistantClient and provide their own customized functionalities?

Could you provide a use case or example?

server/services/assistant_client.ts
Hailong-am: looks good to me.

qianheng-aws: LGTM

gaobinlong (approved):
LGTM.

public/services/assistant_client.ts
```diff
@@ -0,0 +1,35 @@
...
+
+interface Options {
+  dataSourceId?: string;
+}
+
+export class AssistantClient {
+  constructor(private http: HttpSetup) {}
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  executeAgent = (agentId: string, parameters: Record<string, any>, options?: Options) => {
```
qianheng-aws, line 17: Will there be chance that executeAgent and executeAgentByName combined to be one API? Seems the only difference is that they pass different parameters in the `query` field

Otherwise, we need to add active method in both these 2 functions.

ruanyl: this is the APIs of public(UI), they will call the corresponding server APIs, and in server side, they will end up with calling one single function.

PR #313, opened 2024-09-18
feat: use two agents for text2viz

ruanyl:
### Description
Agent 1: rule-based if user didn't provide any instructions
Agent 2: no rules will be applied, only use user's instructions

### Issues Resolved

### Check List

### DIFF
--- a/public/components/visualization/source_selector.tsx
+++ b/public/components/visualization/source_selector.tsx
@@ -14,7 +14,11 @@ import {
   DataSourceOption,
 } from '../../../../../src/plugins/data/public';
 import { StartServices } from '../../types';
-import { TEXT2VEGA_RULE_BASED_AGENT_CONFIG_ID } from '../../../common/constants/llm';
+import {
+  TEXT2PPL_AGENT_CONFIG_ID,
+  TEXT2VEGA_RULE_BASED_AGENT_CONFIG_ID,
+  TEXT2VEGA_WITH_INSTRUCTIONS_AGENT_CONFIG_ID,
+} from '../../../common/constants/llm';
 import { getAssistantService } from '../../services';
 
 const DEFAULT_DATA_SOURCE_TYPE = 'DEFAULT_INDEX_PATTERNS';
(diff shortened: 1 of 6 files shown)


public/components/visualization/source_selector.tsx
```diff
@@ -114,9 +114,12 @@ export const SourceSelector = ({
...
       Object.keys(dataSourceIdToIndexPatternIds).forEach(async (key) => {
-        const res = await assistantService.client.agentConfigExists(TEXT2VEGA_AGENT_CONFIG_ID, {
-          dataSourceId: key !== 'DEFAULT' ? key : undefined,
-        });
+        const res = await assistantService.client.agentConfigExists(
+          TEXT2VEGA_RULE_BASED_AGENT_CONFIG_ID,
+          {
+            dataSourceId: key !== 'DEFAULT' ? key : undefined,
+          }
+        );
```
Hailong-am, line 122: I am not sure whether the logic should be both `TEXT2VEGA_RULE_BASED_AGENT_CONFIG_ID` and `TEXT2VEGA_AGENT_CONFIG_ID` are not configured then disable the index pattern?

ruanyl: Good question, I think both should be configured, or the index pattern should be disabled.
If there is only one configured, it's still problematic, for example, only `TEXT2VEGA_RULE_BASED_AGENT_CONFIG_ID` is configured, but user's input contains instructions, it will use instruction based agent.

Hailong-am: agree, let's do the check for both.

ruanyl: @Hailong-am Updated, the new implementation will check all required agents includes text to ppl, please help to take another look, thanks!

Hailong-am (approved):

ruanyl:
@xluo-aws when merge the PRs, we normally use "Squash and merge" strategy which will create a linear commit history.

PR #327, opened 2024-09-26
refactor(t2viz): register the AI actions to query controls in discover

ruanyl:
### Description
image
image
image

### Issues Resolved

### Check List

### DIFF
--- a/common/types/config.ts
+++ b/common/types/config.ts
@@ -26,7 +26,7 @@ export const configSchema = schema.object({
     enabled: schema.boolean({ defaultValue: false }),
   }),
   branding: schema.object({
-    label: schema.string({ defaultValue: 'AI assistant' }),
+    label: schema.string({ defaultValue: '' }),
   }),
 });
 
--- a/common/types/config.ts
+++ b/common/types/config.ts
@@ -25,6 +25,9 @@ export const configSchema = schema.object({
   smartAnomalyDetector: schema.object({
     enabled: schema.boolean({ defaultValue: false }),
   }),
+  branding: schema.object({
+    label: schema.string({ defaultValue: 'AI assistant' }),
+  }),
 });
 
 export type ConfigSchema = TypeOf<typeof configSchema>;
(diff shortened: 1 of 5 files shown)


common/types/config.ts
```diff
@@ -25,6 +25,9 @@ export const configSchema = schema.object({
   smartAnomalyDetector: schema.object({
     enabled: schema.boolean({ defaultValue: false }),
   }),
+  branding: schema.object({
+    label: schema.string({ defaultValue: 'AI assistant' }),
+  }),
 });
```
Hailong-am, line 31: nit: do we need i18n here?

Hailong-am (approved):

public/components/ui_action_context_menu.tsx
```diff
@@ -38,12 +42,20 @@ export const ActionContextMenu = () => {
   return (
     <EuiPopover
       button={
-        <EuiButtonIcon
+        <EuiButtonEmpty
+          color="text"
           aria-label="AI assistant trigger button"
-          size="s"
-          iconType={assistantTriggerIcon}
+          size="xs"
```
Hailong-am, line 48: maybe we should keep it as `s` to align with DQL

image

common/types/config.ts
ruanyl: This value will be from config yaml, I'm not sure if i18n will even work for config yaml

ruanyl: You mean i18n for the default value?

Hailong-am: > This value will be from config yaml, I'm not sure if i18n will even work for config yaml

I see, if the value from config yaml, the value should be whatever specified in yaml file. can we make default value as empty? that will use the fallback value that supports i18n

public/components/ui_action_context_menu.tsx
ruanyl: having an alignment issue now when changed to a `s` button, this requires a css tweaks in core, I will keep as is now.
image

common/types/config.ts
ruanyl: Good point, I will update

PR #337, opened 2024-09-29
fix: fix create index pattern duplicate cases

raintygao:
### Description
The error of creating duplicate index pattern would not always be the instance of DuplicateIndexPatternError class in some cases, maybe in multi data source or in workspace scenarios, change to finding index pattern directly as a workaround.

NOTE: This is a workaround for now, will find and use another way to fully address after determining the detailed reason. 

### Check List

### DIFF
--- a/public/utils/alerting.ts
+++ b/public/utils/alerting.ts
@@ -48,14 +48,15 @@ export const createIndexPatterns = async (
       dataSourceRef,
     });
   } catch (err) {
-    if (err instanceof DuplicateIndexPatternError) {
+    console.error('Create index pattern error', err.message);
+    // Err instanceof DuplicateIndexPatternError is not a trusted validation in some cases, so find index pattern directly.
+    try {
       const result = await dataStart.indexPatterns.find(patternName);
       if (result && result[0]) {
         pattern = result[0];
       }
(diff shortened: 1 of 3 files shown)


xluo-aws (approved):

public/utils/alerting.ts
```diff
@@ -48,14 +48,15 @@ export const createIndexPatterns = async (
       dataSourceRef,
     });
   } catch (err) {
-    if (err instanceof DuplicateIndexPatternError) {
+    console.error('Create index pattern error', err.message);
+    // Err instanceof DuplicateIndexPatternError is not a trusted validation in some cases, so find index pattern directly.
```
Hailong-am, line 52: out of courisity ,do we have an example of other errors instead of DuplicateIndexPatternError?

Hailong-am (approved):

public/utils/alerting.ts
raintygao: Normally no, because we will create the index pattern according to the agreed parameters, but I guess unless there is a case where the timeFieldName is passed incorrectly, this is the situation we discussed on before that needs to  deep dive alerting DSL mechanism.

PR #343, opened 2024-10-09
BugFix: [AlertInsight] Remove the cache of insight agent id in node server

qianheng-aws:
### Description
Remove the cache of insight agent id in node server.  

~~Pass insight agent id to front-end to indicate that the agent exists, and then the front-end can use the agent id in its next request directly to retrieve insight.~~ 

Detect insight agent id twice to remove the cache:
1. First in summary request to detect whether insight agent exists
2. Second in insight request to get the insight agent id for execution agent.

There will be an potential issue that different requests could be sent to different nodes server and cause error if depending on the cache of insight agent id.

### Check List

### DIFF
--- a/public/components/incontext_insight/generate_popover_body.test.tsx
+++ b/public/components/incontext_insight/generate_popover_body.test.tsx
@@ -87,7 +87,7 @@ describe('GeneratePopoverBody', () => {
         case SUMMARY_ASSISTANT_API.SUMMARIZE:
           value = {
             summary: 'Generated summary content',
-            insightAgentIdExists: true,
+            insightAgentId: 'insightAgentId',
           };
           break;
 
@@ -168,7 +168,7 @@ describe('GeneratePopoverBody', () => {
         case SUMMARY_ASSISTANT_API.SUMMARIZE:
           value = {
             summary: 'Generated summary content',
-            insightAgentIdExists: false,
+            insightAgentId: undefined,
           };
           break;
 
@@ -245,7 +245,7 @@ describe('GeneratePopoverBody', () => {
         case SUMMARY_ASSISTANT_API.SUMMARIZE:
           value = {
             summary: 'Generated summary content',
-            insightAgentIdExists: true,
+            insightAgentId: 'insightAgentId',
           };
           break;
 
@@ -293,7 +293,7 @@ describe('GeneratePopoverBody', () => {
         case SUMMARY_ASSISTANT_API.SUMMARIZE:
           value = {
             summary: 'Generated summary content',
-            insightAgentIdExists: true,
+            insightAgentId: 'insightAgentId',
           };
           break;
 
@@ -335,7 +335,7 @@ describe('GeneratePopoverBody', () => {
         case SUMMARY_ASSISTANT_API.SUMMARIZE:
           value = {
             summary: 'Generated summary content',
-            insightAgentIdExists: true,
+            insightAgentId: 'insightAgentId',
           };
           break;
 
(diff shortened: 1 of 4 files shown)


public/components/incontext_insight/generate_popover_body.test.tsx
```diff
@@ -168,7 +168,7 @@ describe('GeneratePopoverBody', () => {
         case SUMMARY_ASSISTANT_API.SUMMARIZE:
           value = {
             summary: 'Generated summary content',
-            insightAgentIdExists: false,
+            insightAgentId: undefined,
```
yuye-aws, line 171: Why this is undefined and not `insightAgentId`

server/routes/summary_routes.ts
```diff
@@ -57,33 +55,26 @@ export function registerSummaryAssistantRoutes(
...
-          } else if (req.body.insightType === 'user_insight') {
-            if (req.body.type === 'alerts') {
-              if (!userInsightAgentId) {
-                userInsightAgentId = await searchAgent({ name: 'KB_For_Alert_Insight' }, client);
-              }
-            }
-            insightAgentIdExists = !!userInsightAgentId;
+            insightAgentId = await getAgentIdByConfigName(OS_INSIGHT_AGENT_CONFIG_ID, client);
+          } else if (req.body.insightType === 'user_insight' && req.body.type === 'alerts') {
+            insightAgentId = await searchAgent({ name: 'KB_For_Alert_Insight' }, client);
```
yuye-aws, line 66: Shall we store opensearch kb agent id and customer kb agent id into two separate variables? I wonder whether there is some override issue.

public/components/incontext_insight/generate_popover_body.test.tsx
qianheng-aws: `insightAgentId: undefined`  means no insight agentId for this alert, which is semantic equal to `insightAgentIdExists: false`

server/routes/summary_routes.ts
qianheng-aws: It doesn't store the agent id but only pass it to front-end

public/components/incontext_insight/generate_popover_body.test.tsx
yuye-aws: Makes sense

server/routes/summary_routes.ts
yuye-aws: Makes sense

server/routes/summary_routes.ts
```diff
@@ -128,15 +114,25 @@ export function registerSummaryAssistantRoutes(
         return res.ok({ body: response.body.inference_results[0].output[0].result });
       } catch (e) {
         return res.internalError();
```
Hailong-am, line 116: not related to this change, but we don't return 500 status code

server/routes/summary_routes.ts
```diff
@@ -128,15 +114,25 @@ export function registerSummaryAssistantRoutes(
...
+  summaryType: string,
+  client: OpenSearchClient['transport']
+) {
+  // We have separate agent for os_insight and user_insight. And for user_insight, we can
+  // only get it by searching on name since it is not stored in agent config.
+  if (insightType === 'os_insight') {
+    return getAgentIdByConfigName(OS_INSIGHT_AGENT_CONFIG_ID, client);
+  } else if (insightType === 'user_insight' && summaryType === 'alerts') {
+    return searchAgent({ name: 'KB_For_Alert_Insight' }, client);
+  }
```
Hailong-am, line 133: missing return value if `insightType` is not `os_insight` or `user_insight`

qianheng-aws: Seems it will return undefined implicitly in that case(I tell that by seeing the return type in my IDE). But it's still better to return undefined explicitly.

server/routes/summary_routes.ts
qianheng-aws: Which status code should we return?

ruanyl (approved):

Hailong-am (approved):

PR #349, opened 2024-10-15
feat: take index pattern and user input to t2viz from discover

ruanyl:
### Description
[Describe what this change achieves]

https://github.com/user-attachments/assets/2ba696b2-2f34-4dc9-8093-a3e55a1e7fa4

### Issues Resolved

### Check List

### DIFF
--- a/public/plugin.tsx
+++ b/public/plugin.tsx
@@ -60,6 +60,10 @@ import {
 } from './vis_nlq/saved_object_loader';
 import { NLQVisualizationEmbeddableFactory } from './components/visualization/embeddable/nlq_vis_embeddable_factory';
 import { NLQ_VISUALIZATION_EMBEDDABLE_TYPE } from './components/visualization/embeddable/nlq_vis_embeddable';
+import {
+  ASSISTANT_INPUT_URL_SEARCH_KEY,
+  INDEX_PATTERN_URL_SEARCH_KEY,
+} from './components/visualization/text2viz';
 
 export const [getCoreStart, setCoreStart] = createGetterSetter<CoreStart>('CoreStart');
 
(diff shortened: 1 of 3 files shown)


public/plugin.tsx
```diff
@@ -329,7 +333,24 @@ export class AssistantPlugin
         getDisplayName: () => 'Generate visualization',
         getIconType: () => 'visLine' as const,
         execute: async () => {
-          core.application.navigateToApp(TEXT2VIZ_APP_ID);
+          const dataset = data.query.queryString.getQuery().dataset;
+          const url = new URL(core.application.getUrlForApp(TEXT2VIZ_APP_ID, { absolute: true }));
+          if (dataset && dataset.type === 'INDEX_PATTERN') {
+            url.searchParams.set(INDEX_PATTERN_URL_SEARCH_KEY, dataset.id);
```
gaobinlong, line 339: Don't we need the data source info?

ruanyl: Not for t2viz at the moment, because it will request the index pattern details by id, and retrieve data source id from there.

public/plugin.tsx
```diff
@@ -329,7 +333,24 @@ export class AssistantPlugin
         getDisplayName: () => 'Generate visualization',
         getIconType: () => 'visLine' as const,
         execute: async () => {
-          core.application.navigateToApp(TEXT2VIZ_APP_ID);
+          const dataset = data.query.queryString.getQuery().dataset;
+          const url = new URL(core.application.getUrlForApp(TEXT2VIZ_APP_ID, { absolute: true }));
+          if (dataset && dataset.type === 'INDEX_PATTERN') {
```
Hailong-am, line 338: nit: we can replace the hardcoded value with `https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/common/constants.ts#L53`

public/plugin.tsx
```diff
@@ -329,7 +333,24 @@ export class AssistantPlugin
...
+           * once query assistant is moved to dashboard-assistant repo. Currently, there is no better
+           * way to get the input value as the query assistant is currently implemented in OSD core.
+           */
+          const queryAssistInputEle = document.getElementsByClassName('queryAssist__input')[0];
+          if (queryAssistInputEle instanceof HTMLInputElement) {
+            const input = queryAssistInputEle.value;
+            if (input) {
+              url.searchParams.set(ASSISTANT_INPUT_URL_SEARCH_KEY, input);
+            }
+          }
```
Hailong-am, line 352: can we get the value from the place triggered this action? If so, we can use context to pass the value.

Hailong-am (approved):

public/plugin.tsx
ruanyl: It's a bit tricky, the context menu is updated based on `data.query.queryString.getUpdates$()`, but the query assistant input is not part of it. Didn't find a way to easily update the context when the input changes. I have a todo to refactor this when we moved query assistant to dashboards-assistant plugin.

PR #376, opened 2024-12-13
Refactor multi agents execution

ruanyl:
### Description
The refactor in the commit introduces a `Pipeline` to execute asynchronous operations, each operator encapsulates a specific function. The Pipeline orchestrates these operators, processing input sequentially. The pipeline architecture allows for adding, reusing or rearranging steps with minimal impact on the codebase.

The monolithic Text2Vega class also be replaced with Pipeline in this PR.

### Issues Resolved

### Check List


public/utils/pipeline/operator.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export abstract class Operator<T, P> {
+  abstract execute(v: T): Promise<P>;
```
Hailong-am, line 7: do we need context parameter? and how about name it as `Task`

ruanyl: > how about name it as Task

Sounds good, I like the name, let me update.

> do we need context parameter?

Is there an example?

Hailong-am: I don't have a real use case, but this could helpful if we have context for tasks share intermediate result during execution. For example, task 3 want to use a output of task 1

Hailong-am (approved):

public/utils/pipeline/operator.ts
ruanyl: The current design is each task will add extra properties to its output, and the output will become the next task's input. But I got your point which makes to me, I think we can tweaks the implementation when there is a real use case.

PR #373, opened 2024-11-19
Optimize the response of AI agent APIs

gaobinlong:
### Description

When calling the APIs like api/assistant/summary, api/assistant/insight and other related APIs, if OpenSearch cluster returns 4xx error, we just return 500 error or 400 error directly, which is not elegant, we should optimize this by checking the status code, if it's 4xx error, we return the original response, and if it's 5xx error, return an general error message.

### Issues Resolved
[List any issues this PR will resolve]

### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass, including unit test, integration test.
- [ ] New functionality has user manual doc added.
- [ ] Commits are signed per the DCO using --signoff.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #380, opened 2024-12-26
Fix returns 500 error for AI agent APIs when OpenSearch returns 4xx error with json format error message

gaobinlong:
### Description
In this PR:https://github.com/opensearch-project/dashboards-assistant/pull/373, we returns 4xx error for the AI agent APIs when the execute agent API in ml-commons plugin responds with a 4xx error, but previously, the execute agent API in ml-commons plugin returns string format error message if the request failed, now it changed to json format, this triggers different condition [here](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/6e851ff33d155e39734707b49a62be4f13ece6ee/src/core/server/http/router/response_adapter.ts#L188) in OSD-core, so we the AI agent APIs return 500 internal error because the custom response body doesn't have a `message` field, this PR fixed the bug.

In addition, when customizing the response, `headers` is not needed because the headers comes from ml-commons API, that is useless for node APIs.

image

image

### Issues Resolved
No issue.

### Check List

### DIFF
--- a/server/routes/summary_routes.ts
+++ b/server/routes/summary_routes.ts
@@ -47,7 +47,6 @@ export function registerSummaryAssistantRoutes(
         req.body.index && req.body.dsl && req.body.topNLogPatternData
           ? LOG_PATTERN_SUMMARY_AGENT_CONFIG_ID
           : SUMMARY_AGENT_CONFIG_ID;
-
       let response;
       try {
         response = await assistantClient.executeAgentByConfigName(agentConfigId, {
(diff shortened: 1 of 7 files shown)


ruanyl (approved):
Looks good! Thank you for adding the tests

server/routes/summary_routes.ts
```diff
@@ -61,15 +60,13 @@ export function registerSummaryAssistantRoutes(
...
           return res.customError({
-            body: e.body,
+            body: { message: typeof e.body === 'string' ? e.body : JSON.stringify(e.body) },
             statusCode: e.statusCode,
-            headers: e.headers,
           });
         } else {
           return res.customError({
             body: 'Execute agent failed!',
             statusCode: 500,
```
xluo-aws, line 69: Should we return original statusCode? As long as we don't disclose server side information, we should be fine.

ruanyl:
I will merge now, will address @xluo-aws concern in a separate PR

gaobinlong:
> I will merge now, will address @xluo-aws concern in a separate PR

@xluo-aws @ruanyl OK, I'll do some test to verify that return the original 5xx status code doesn't introduce some unexpected behavior.

PR #394, opened 2025-01-20
Hide "stop generation" and regenerate button based on feature flag

hutiechuan:
### Description
[Describe what this change achieves]

### Issues Resolved

### Check List

### DIFF
--- a/common/types/config.ts
+++ b/common/types/config.ts
@@ -13,6 +13,7 @@ export const configSchema = schema.object({
     feedback: schema.boolean({ defaultValue: true }),
     allowRenameConversation: schema.boolean({ defaultValue: true }),
     deleteConversation: schema.boolean({ defaultValue: true }),
+    hideRegenerateButton: schema.boolean({ defaultValue: true }),
   }),
   incontextInsight: schema.object({
     enabled: schema.boolean({ defaultValue: true }),
--- a/common/types/config.ts
+++ b/common/types/config.ts
@@ -13,7 +13,7 @@ export const configSchema = schema.object({
     feedback: schema.boolean({ defaultValue: true }),
     allowRenameConversation: schema.boolean({ defaultValue: true }),
     deleteConversation: schema.boolean({ defaultValue: true }),
-    regenerateButton: schema.boolean({ defaultValue: true }),
+    regenerateMessage: schema.boolean({ defaultValue: true }),
   }),
   incontextInsight: schema.object({
     enabled: schema.boolean({ defaultValue: true }),
--- a/common/types/config.ts
+++ b/common/types/config.ts
@@ -13,7 +13,7 @@ export const configSchema = schema.object({
     feedback: schema.boolean({ defaultValue: true }),
     allowRenameConversation: schema.boolean({ defaultValue: true }),
     deleteConversation: schema.boolean({ defaultValue: true }),
-    hideRegenerateButton: schema.boolean({ defaultValue: true }),
+    regenerateButton: schema.boolean({ defaultValue: true }),
   }),
   incontextInsight: schema.object({
     enabled: schema.boolean({ defaultValue: true }),
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "eslint": "^6.8.0",
     "husky": "^8.0.0",
     "jest-dom": "^4.0.0",
-    "lint-staged": "^10.2.0",
+    "lint-staged": "^13.1.0",
     "ts-jest": "^29.1.0"
   },
   "eslintIgnore": [
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "eslint": "^6.8.0",
     "husky": "^8.0.0",
     "jest-dom": "^4.0.0",
-    "lint-staged": "^13.1.0",
+    "lint-staged": "^10.2.0",
     "ts-jest": "^29.1.0"
   },
   "eslintIgnore": [
--- a/public/tabs/chat/chat_page_content.tsx
+++ b/public/tabs/chat/chat_page_content.tsx
@@ -30,7 +30,6 @@ interface ChatPageContentProps {
   messagesLoading: boolean;
   messagesLoadingError?: Error;
   onRefresh: () => void;
-  hideRegenerateButton?: boolean;
 }
 
 export const ChatPageContent: React.FC<ChatPageContentProps> = React.memo((props) => {
@@ -41,7 +40,7 @@ export const ChatPageContent: React.FC<ChatPageContentProps> = React.memo((props
   const chatActions = useChatActions();
   const registry = getIncontextInsightRegistry();
   const configSchema = getConfigSchema();
-  const hideRegenerateButton = configSchema.chat.hideRegenerateButton ?? false;
+  const showRegenerate = configSchema.chat.regenerateButton;
 
   useLayoutEffect(() => {
     pageEndRef.current?.scrollIntoView();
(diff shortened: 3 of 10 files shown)


public/tabs/chat/messages/message_action.tsx
```diff
@@ -107,7 +109,7 @@ export const MessageActions: React.FC<MessageActionsProps> = ({
       ),
     },
     regenerate: {
-      show: showRegenerate && onRegenerate,
+      show: hideRegenerateButton && showRegenerate && onRegenerate,
```
lyx-aws, line 112: Should this be `!hideRegenerateButton`?

package.json
```diff
@@ -33,7 +33,7 @@
     "eslint": "^6.8.0",
     "husky": "^8.0.0",
     "jest-dom": "^4.0.0",
-    "lint-staged": "^13.1.0",
+    "lint-staged": "^10.2.0",
```
lyx-aws, line 36: Why are we changing this?

public/tabs/chat/messages/message_action.tsx
hutiechuan: yes

yarn.lock
ruanyl, line 1: Seems this file change is not needed, we can revert this.

public/tabs/chat/chat_page_content.tsx
```diff
@@ -144,7 +149,8 @@ export const ChatPageContent: React.FC<ChatPageContentProps> = React.memo((props
           <MessageBubble loading showActionBar={false} />
         </>
       )}
-      {chatState.llmResponding && chatContext.conversationId && (
+
+      {hideRegenerateButton && chatState.llmResponding && chatContext.conversationId && (
```
lyx-aws, line 153: Should this also be `!hideRegenerateButton`?

public/tabs/chat/chat_page_content.tsx
```diff
@@ -24,10 +24,13 @@ import { MessageContent } from './messages/message_content';
 import { SuggestionBubble } from './suggestions/suggestion_bubble';
 import { getIncontextInsightRegistry } from '../../services';
 
+import { getConfigSchema } from '../../services';
+
 interface ChatPageContentProps {
   messagesLoading: boolean;
   messagesLoadingError?: Error;
   onRefresh: () => void;
+  hideRegenerateButton?: boolean;
```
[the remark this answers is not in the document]
hutiechuan: ok

public/tabs/chat/messages/message_action.tsx
```diff
@@ -31,6 +31,7 @@ interface MessageActionsProps {
   usageCollection?: UsageCollectionSetup;
   metricAppName?: string;
   buttonOrder?: string[];
+  hideRegenerateButton?: boolean;
```
ruanyl, line 34: We already have `showRegenerate` props, can we leverage that?

hutiechuan: yes i will do that

hutiechuan: showRegenerate is already in the codebase, but that is used to explicitly control whether the regenerate button is shown; it doesn't get the value from config.

hutiechuan: basically showRegenerate props is being passed in the chat_page_content.tsx.

the showRegenerate value is set based on whether the current message is the latest output after the user's input, which is doing the different thing as hideRegenerateButton does.

ruanyl: I was thinking in `chat_page_content.tsx`, we can do:
```
showRegenerate={isLatestOutput && !configSchema.chat.hideRegenerateButton}
```
because for `MessageBubble` component, it doesn't need to know the logic of how to handle the visibility of regenerate button.

hutiechuan: yeah, i think that makes sense

hutiechuan: because showRegenerate code was there before I modified anything.
I saw the logic there, I assumed the previous coder might needs variable showRegenerate for a specific reason. So i just added one more variable hideRegenerateButton in case i didn't mess anything up

package.json
ruanyl: Let's revert this change, if you run `yarn osd bootstrap --single-version=loose`, it should be good.

hutiechuan: ok

common/types/config.ts
```diff
@@ -11,6 +11,7 @@ export const configSchema = schema.object({
     enabled: schema.boolean({ defaultValue: false }),
     trace: schema.boolean({ defaultValue: true }),
     feedback: schema.boolean({ defaultValue: true }),
+    hideRegenerateButton: schema.boolean({ defaultValue: true }),
```
SuZhou-Joe, line 14:
```suggestion
    regenerateButton: schema.boolean({ defaultValue: true }),
```

hutiechuan: Yes.
There isn't regenerateButton initially.
Do you want me to delete hideRegenerateButton and create a new variable called regenerateButton?

SuZhou-Joe: To keep the convention, it seems we use positive value by default for each feature, and we have to explicitly `turn off` the features if we do not need it.

For the previous code, by default the regenerate button will be disabled by default, which is not expected.

hutiechuan: ok, I see. Just to confirm with you, basically i need to rename the variable from hideRegenerateButton to regenerateButton to follow the convention right?

SuZhou-Joe: Yeah, I think so.

hutiechuan: so i guess previous i was written show: !hideregenerateButton && showRegenerate && onRegenerate,
it should be show: regenerateButton && showRegenerate && onRegenerate, and so on for other examples

hutiechuan: so i guess previously i was written show: !hideregenerateButton && showRegenerate && onRegenerate,
it should now be show: regenerateButton && showRegenerate && onRegenerate, and so on for other examples

SuZhou-Joe: Yes.

hutiechuan: done

common/types/config.ts
```diff
@@ -11,6 +11,7 @@ export const configSchema = schema.object({
     enabled: schema.boolean({ defaultValue: false }),
     trace: schema.boolean({ defaultValue: true }),
     feedback: schema.boolean({ defaultValue: true }),
+    regenerateButton: schema.boolean({ defaultValue: true }),
```
ruanyl, line 14: I suggest rename it to `regenerateMessage` to avoid mentioning `Button` in the flag, because the UI might be different in the future.

public/tabs/chat/chat_page_content.tsx
```diff
@@ -37,6 +39,8 @@ export const ChatPageContent: React.FC<ChatPageContentProps> = React.memo((props
   const loading = props.messagesLoading || chatState.llmResponding;
   const chatActions = useChatActions();
   const registry = getIncontextInsightRegistry();
+  const configSchema = getConfigSchema();
+  const showRegenerate = configSchema.chat.regenerateButton;
```
ruanyl, line 43: If we do this on line 133, then no changes are required for `message_bubble.tsx` and `message_action.tsx`
```
showRegenerate={isLatestOutput && showRegenerate}
```

ruanyl (approved):

PR #400, opened 2025-01-21
disable rename conversation api using feature flag

kamahuan:
### Description
Disable the rename conversation api using feature flag. Upon checking with Binlong and Arjun, disable the api from the frontend is sufficient.

### Issues Resolved
NONE

### Check List

### DIFF
--- a/common/types/config.ts
+++ b/common/types/config.ts
@@ -11,6 +11,7 @@ export const configSchema = schema.object({
     enabled: schema.boolean({ defaultValue: false }),
     trace: schema.boolean({ defaultValue: true }),
     feedback: schema.boolean({ defaultValue: true }),
+    allowRenameConversation: schema.boolean({ defaultValue: true }),
   }),
   incontextInsight: schema.object({
     enabled: schema.boolean({ defaultValue: true }),
--- a/public/components/edit_conversation_name_modal.tsx
+++ b/public/components/edit_conversation_name_modal.tsx
@@ -8,6 +8,7 @@ import React, { useCallback, useRef } from 'react';
 import { EuiConfirmModal, EuiCompressedFieldText, EuiSpacer, EuiText } from '@elastic/eui';
 import { useCore } from '../contexts/core_context';
 import { usePatchConversation } from '../hooks';
+import { getConfigSchema } from '../services';
 
 export interface EditConversationNameModalProps {
   onClose?: (status: 'updated' | 'cancelled' | 'errored', newTitle?: string) => void;
@@ -25,6 +26,7 @@ export const EditConversationNameModal = ({
       notifications: { toasts },
     },
   } = useCore();
+  const configSchema = getConfigSchema();
   const titleInputRef = useRef<HTMLInputElement>(null);
   const { loading, abort, patchConversation, isAborted } = usePatchConversation();
 
@@ -50,6 +52,13 @@ export const EditConversationNameModal = ({
     }
     onClose?.('updated', title);
   }, [onClose, conversationId, patchConversation, toasts.addSuccess, toasts.addDanger, isAborted]);
+  console.log('Rename conversation enabled:', configSchema.chat.allowRenameConversation);
+
+  // Move the feature flag check after all hooks are declared
+  if (!configSchema.chat.allowRenameConversation) {
+    onClose?.('cancelled');
+    return null;
+  }
 
   return (
     <EuiConfirmModal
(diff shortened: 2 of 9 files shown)


public/components/edit_conversation_name_modal.tsx
```diff
@@ -50,6 +52,13 @@ export const EditConversationNameModal = ({
     }
     onClose?.('updated', title);
   }, [onClose, conversationId, patchConversation, toasts.addSuccess, toasts.addDanger, isAborted]);
+  console.log('Rename conversation enabled:', configSchema.chat.allowRenameConversation);
```
ruanyl, line 55: Let's cleanup console.log :)

common/types/config.ts
```diff
@@ -11,6 +11,7 @@ export const configSchema = schema.object({
     enabled: schema.boolean({ defaultValue: false }),
     trace: schema.boolean({ defaultValue: true }),
     feedback: schema.boolean({ defaultValue: true }),
+    allowRenameConversation: schema.boolean({ defaultValue: true }),
```
ruanyl, line 14: What about name it as `coversationManagement`? This flag will be used to control conversation edit and delete, am I right?

public/tabs/history/chat_history_list.tsx
```diff
@@ -74,14 +78,16 @@ export const ChatHistoryListItem = ({
...
+              <EuiFlexItem grow={false}>
+                <EuiSmallButtonIcon
+                  onClick={handleEditClick}
+                  iconType="pencil"
+                  aria-label="Edit conversation name"
+                  color="text"
+                />
+              </EuiFlexItem>
+            )}
             <EuiFlexItem grow={false}>
```
ruanyl, line 91: we also need to hide the delete button.

common/types/config.ts
kamahuan: This flag and PR is only about disabling the rename conversation feature. Will submit another PR about disabling delete conversation api.

public/tabs/history/chat_history_list.tsx
kamahuan: Will submit a separate PR about deleting the conversation api.

public/components/edit_conversation_name_modal.tsx
```diff
@@ -50,6 +52,13 @@ export const EditConversationNameModal = ({
...
+
+  // Move the feature flag check after all hooks are declared
+  if (!configSchema.chat.allowRenameConversation) {
+    onClose?.('cancelled');
+    return null;
```
ruanyl, line 60: Q: curious why adding this? My understanding is there will be no place to trigger this modal as we hide the edit button from the dropdown menu. Perhaps not necessary?

public/tabs/history/__tests__/chat_history_list.test.tsx
```diff
@@ -7,9 +7,28 @@ import React from 'react';
...
+        feedback: true,
+      },
+      incontextInsight: { enabled: true },
+      next: { enabled: false },
+      text2viz: { enabled: false },
+      alertInsight: { enabled: false },
+      smartAnomalyDetector: { enabled: false },
+      branding: { label: undefined, logo: undefined },
+    });
+
```
ruanyl, line 31: I see this piece of code is repeating in many test files, what about to have this mock in `test/setup.jest.ts`? This will make the mock work every tests, and we only need to do manual mock at where we need to test the change the default value.


// test/setup.jest.ts
```
const DEFAULT_CONFIG = Object.freeze({
  enabled: true,
  chat: {
    enabled: true,
    allowRenameConversation: true,
    trace: true,
    feedback: true,
  },
  incontextInsight: { enabled: true },
  next: { enabled: false },
  text2viz: { enabled: false },
  alertInsight: { enabled: false },
  smartAnomalyDetector: { enabled: false },
  branding: { label: undefined, logo: undefined },
});
jest.spyOn(services, 'getConfigSchema').mockReturnValue(DEFAULT_CONFIG);
```

common/types/config.ts
ruanyl: I see, that also works :)

ruanyl:
@kamahuan would you mind to change the target branch to `main` and add an entry in CHANGELOG.md? That will make the backport a bit easier :)

public/components/edit_conversation_name_modal.tsx
kamahuan: Good catch. we can remove this feature flag check and test since the edit button is already hidden when the feature is disabled. Having the check in both places is redundant. I'll update the PR to simplify this.

public/tabs/history/__tests__/chat_history_list.test.tsx
kamahuan: Yes will use a separate file to put this config

public/components/edit_conversation_name_modal.tsx
kamahuan: Removed

PR #401, opened 2025-01-22
feat: Set logo config in assistant and read logo by config

raintygao:
### Description
Set logo config in assistant and read logo by config

### Screenshot
image
image
image
image

### Check List

### DIFF
--- a/public/chat_header_button.tsx
+++ b/public/chat_header_button.tsx
@@ -26,6 +26,7 @@ import {
 import { useCore } from './contexts/core_context';
 import { MountPointPortal } from '../../../src/plugins/opensearch_dashboards_react/public';
 import { usePatchFixedStyle } from './hooks/use_patch_fixed_style';
+import { getConfigSchema } from './services';
 
 interface HeaderChatButtonProps {
   application: ApplicationStart;
@@ -49,7 +50,7 @@ export const HeaderChatButton = (props: HeaderChatButtonProps) => {
   const [inputFocus, setInputFocus] = useState(false);
   const inputRef = useRef<HTMLInputElement>(null);
   const registry = getIncontextInsightRegistry();
-
+  const config = getConfigSchema();
   const [sidecarDockedMode, setSidecarDockedMode] = useState(DEFAULT_SIDECAR_DOCKED_MODE);
   const core = useCore();
   const flyoutFullScreen = sidecarDockedMode === SIDECAR_DOCKED_MODE.TAKEOVER;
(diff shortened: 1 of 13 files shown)


public/components/incontext_insight/index.tsx
```diff
@@ -275,7 +282,7 @@ export const IncontextInsight = ({
         </EuiFlexItem>
         <EuiFlexItem grow={1}>
           <div className="incontextInsightAnchorIcon">
-            <EuiIcon type={input.type === 'generate' ? sparkleIcon : chatIcon} size="l" />
+            <EuiIcon type={logoIcon} size="m" />
```
wanglam, line 285: Do we need to use different icon for generate?

public/chat_header_button.tsx
```diff
@@ -61,6 +62,10 @@ export const HeaderChatButton = (props: HeaderChatButtonProps) => {
     return () => subscription.unsubscribe();
   });
 
+  const logoIcon = useMemo(() => {
+    return config?.branding?.logo?.gray ?? chatIcon;
+  }, [config]);
```
wanglam, line 67: Nit: Shall we separate a icon util or hook to fetch these icons? I'm not sure if config change require page reload. If yes, I think we don't need to `useMemo` here. 
Separate a util like below would be more convenient.
```tsx
const getLogoIcon = (type: "gray"|"gradient", fallback=chatIcon)
```

public/components/incontext_insight/index.tsx
raintygao: No, they are both needed to be replaced with latest.

public/chat_header_button.tsx
raintygao: Sounds good, updated.

public/components/incontext_insight/generate_popover_body.tsx
```diff
@@ -297,7 +296,12 @@ export const GeneratePopoverBody: React.FC<{
                 color={'text'}
               />
             ) : (
-              <EuiIcon aria-label="alert-assistant" color="hollow" size="l" type={shiny_sparkle} />
+              <EuiIcon
+                aria-label="alert-assistant"
+                color="hollow"
+                size="m"
+                type={getLogoIcon('gradient')}
```
wanglam, line 303: Is it safe to use `chatIcon` as the logo icon fallback? It's using `shiny_sparkle` before.

raintygao: I think so because shiny_sparkle should also be updated to latest icon, but I would change to keep shiny_sparkle as default because it looks better than raw chatIcon.

wanglam (approved):

ruanyl (approved):

PR #417, opened 2025-01-27
Hide the assistant entry when there isn't data2summary agent

Qxisylolo:
### Description
This pr hides the assistant entry when there isn't data2summary agent
!! This pr is a hard dependency that should be merged after https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9277 has been merged.

has agent:
截屏2025-01-27 13 45 13
no agent:
截屏2025-01-27 13 44 44

### Check List


public/components/ui_action_context_menu.tsx
```diff
@@ -39,6 +42,35 @@ export const ActionContextMenu = (props: Props) => {
...
+    return response;
+  }
+
+  useEffect(() => {
+    const fetchSummaryAgent = async () => {
+      try {
+        const summaryAgentStatus = await checkAgentsExist(
+          props.httpSetup,
+          DATA2SUMMARY_AGENT_CONFIG_ID,
+          props.data.query.queryString.getQuery().dataset?.dataSource?.id
```
SuZhou-Joe, line 64: We should listen to dataSourceId change. And if the resultSummaryEnabled is false, we do not need to check if agents exists.

public/components/ui_action_context_menu.tsx
```diff
@@ -39,6 +42,35 @@ export const ActionContextMenu = (props: Props) => {
...
+        );
+        props.isSummaryAgentAvailable$.next(!!summaryAgentStatus.exists);
+      } catch (error) {
+        console.error(error);
+      }
+    };
+
+    fetchSummaryAgent();
+  }, [props, props.httpSetup]);
```
SuZhou-Joe, line 73: We should listen to dataSourceId change. And if the resultSummaryEnabled is false, we do not need to check if agents exists.

public/components/ui_action_context_menu.tsx
```diff
@@ -75,14 +107,13 @@ export const ActionContextMenu = (props: Props) => {
     [actionContext.datasetId, actionContext.datasetType, actionContext.dataSourceId]
   );
 
-  // The action button should be not displayed when there is no action and result summary disabled.
-  if (actionsRef.current.length === 0 && !resultSummaryEnabled) {
+  // The action button should be not displayed when there is no action and result summary disabled or there is no data2Summary agent
+  if (!isSummaryAgentAvailable || (actionsRef.current.length === 0 && !resultSummaryEnabled)) {
```
SuZhou-Joe, line 111: I'd prefer to move isSummaryAgentAvailable behind !resultSummaryEnabled because we need to make a network request to get the status.

public/plugin.tsx
```diff
@@ -282,7 +282,11 @@ export class AssistantPlugin
       }
     })();
 
-    const { isQuerySummaryCollapsed$, resultSummaryEnabled$ } = setupDeps.queryEnhancements;
+    const {
+      isQuerySummaryCollapsed$,
+      resultSummaryEnabled$,
+      isSummaryAgentAvailable$,
```
SuZhou-Joe, line 288: Do we have such observable from queryEnhancements plugin?

Qxisylolo: I add one in osd repo

public/components/ui_action_context_menu.tsx
```diff
@@ -39,6 +42,35 @@ export const ActionContextMenu = (props: Props) => {
...
+  async function checkAgentsExist(
+    http: HttpSetup,
+    agentConfigName: string | string[],
+    dataSourceId?: string
+  ) {
+    const response = await http.get(AGENT_API.CONFIG_EXISTS, {
+      query: { agentConfigName, dataSourceId },
+    });
+    return response;
+  }
```
SuZhou-Joe, line 56: We should move this function out of render function.

public/plugin.tsx
SuZhou-Joe: Thanks, then we'd better mention that in the PR and it is a hard dependency that this PR should not merge until that PR gets merged.

public/components/ui_action_context_menu.tsx
```diff
@@ -75,14 +107,13 @@ export const ActionContextMenu = (props: Props) => {
...
-  // The action button should be not displayed when there is no action and result summary disabled.
-  if (actionsRef.current.length === 0 && !resultSummaryEnabled) {
+  // The action button should be not displayed when there is no action and result summary disabled or there is no data2Summary agent
+  if (!isSummaryAgentAvailable || (actionsRef.current.length === 0 && !resultSummaryEnabled)) {
     return null;
   }
 
-  // The action button should be disabled when context menu has no item and result summary disabled.
+  // The action button should be disabled when context menu has no item and result summary disabled
   const actionDisabled = (panels.value?.[0]?.items ?? []).length === 0 && !resultSummaryEnabled;
```
SuZhou-Joe, line 116:
```suggestion
  const actionDisabled = (panels.value?.[0]?.items ?? []).length === 0 && !resultSummaryEnabled && !isSummaryAgentAvailable;
```
Do we need to add check on `isSummaryAgentAvailable`?

public/plugin.tsx
Qxisylolo: ok, I have added it, thank you

public/components/ui_action_context_menu.tsx
Qxisylolo: I add this check at first, but if there isn't agent, should we just hide the button instead of disabling it?

SuZhou-Joe: I think it is OK that we check on both places, but sure it is good enough if we will hide the button if no summary agent is available on target data source.

public/components/ui_action_context_menu.tsx
```diff
@@ -39,6 +42,35 @@ export const ActionContextMenu = (props: Props) => {
...
+    const response = await http.get(AGENT_API.CONFIG_EXISTS, {
+      query: { agentConfigName, dataSourceId },
+    });
+    return response;
+  }
+
+  useEffect(() => {
+    const fetchSummaryAgent = async () => {
+      try {
+        const summaryAgentStatus = await checkAgentsExist(
```
SuZhou-Joe, line 61: Do we need to `props.isSummaryAgentAvailable$.next(false)` before making the call to check if agent exists? if we switch from data source A(agent configured) to data source B (agent not configured) and the call somehow failed, the status will be incorrect.

public/components/ui_action_context_menu.tsx
Qxisylolo: update~ and thanks for all the comments

public/components/ui_action_context_menu.tsx
```diff
@@ -75,13 +99,15 @@ export const ActionContextMenu = (props: Props) => {
     [actionContext.datasetId, actionContext.datasetType, actionContext.dataSourceId]
   );
 
-  // The action button should be not displayed when there is no action and result summary disabled.
-  if (actionsRef.current.length === 0 && !resultSummaryEnabled) {
+  // The action button should be not displayed when there is no action and result summary disabled or there is no data2Summary agent
+  if ((!resultSummaryEnabled && actionsRef.current.length === 0) || !isSummaryAgentAvailable) {
```
wanglam, line 103: Shall we need to allow other actions to be displayed if the summary agent is not available? The other actions won't be displayed if the summary agent is not available after the current change. Is it possible to not register the summary action if the summary agent is not available?

xluo-aws: The other action includes anomaly detection suggestion and generate visualization, right? If yes, They should be displayed even if their agents are configured, should not be impacted by summary agent.

Qxisylolo: Thanks for the comments, already updated, now the UI context bar will be disabled only when there isn't actions and result summary disabled or or no data2Summary agent is available

public/components/ui_action_context_menu.tsx
```diff
@@ -39,6 +43,26 @@ export const ActionContextMenu = (props: Props) => {
   });
   const resultSummaryEnabled = useObservable(props.resultSummaryEnabled$, false);
   const isQuerySummaryCollapsed = useObservable(props.isQuerySummaryCollapsed$, false);
+  const isSummaryAgentAvailable = useObservable(props.isSummaryAgentAvailable$, false);
```
wanglam, line 46: I prefer to add a `shouldShowSummarizationAction` flag here. Something like
```ts
const shouldShowSummarizationAction = resultSummaryEnabled && isSummaryAgentAvailable;
```
We always need to check `resultSummaryEnabled` and `isSummaryAgentAvailable` when we need to determine if this action can be displayed. I'm suggest to combine them in same flag. It can improve the code readability.

public/components/ui_action_context_menu.tsx
```diff
@@ -75,13 +99,15 @@ export const ActionContextMenu = (props: Props) => {
     [actionContext.datasetId, actionContext.datasetType, actionContext.dataSourceId]
   );
 
-  // The action button should be not displayed when there is no action and result summary disabled.
-  if (actionsRef.current.length === 0 && !resultSummaryEnabled) {
+  // The action button should be not displayed when there is no action and result summary disabled or there is no data2Summary agent
+  if ((!resultSummaryEnabled || !isSummaryAgentAvailable) && actionsRef.current.length === 0) {
```
wanglam, line 103: Can be replace with `actionsRef.current.length === 0 && !shouldShowSummarizationAction`

public/components/ui_action_context_menu.tsx
```diff
@@ -75,13 +99,15 @@ export const ActionContextMenu = (props: Props) => {
...
     return null;
   }
 
-  // The action button should be disabled when context menu has no item and result summary disabled.
-  const actionDisabled = (panels.value?.[0]?.items ?? []).length === 0 && !resultSummaryEnabled;
```
wanglam, line 84: Can be replace with `(panels.value?.[0]?.items ?? []).length === 0 && ! shouldShowSummarizationAction`.

public/components/ui_action_context_menu.tsx
wanglam, line 1: Should replace `resultSummaryEnabled` with `shouldShowSummarizationAction` in line 138.

public/components/ui_action_context_menu.tsx
Qxisylolo: yes, it's a good advice, update~

public/components/ui_action_context_menu.tsx
```diff
@@ -39,6 +43,27 @@ export const ActionContextMenu = (props: Props) => {
   });
   const resultSummaryEnabled = useObservable(props.resultSummaryEnabled$, false);
   const isQuerySummaryCollapsed = useObservable(props.isQuerySummaryCollapsed$, false);
+  const isSummaryAgentAvailable = useObservable(props.isSummaryAgentAvailable$, false);
+  const shouldShowSummarizationAction = resultSummaryEnabled && isSummaryAgentAvailable;
+
+  useEffect(() => {
+    props.isSummaryAgentAvailable$.next(false);
+    if (!resultSummaryEnabled) return;
```
Hailong-am, line 51: nit: change the order of these two lines may reduce the times `props.isSummaryAgentAvailable$` emit values that may cause re-rendering when `resultSummaryEnabled` is false

public/components/ui_action_context_menu.tsx
```diff
@@ -39,6 +43,27 @@ export const ActionContextMenu = (props: Props) => {
...
+    const fetchSummaryAgent = async () => {
+      try {
+        const summaryAgentStatus = await checkAgentsExist(
+          props.httpSetup,
+          DATA2SUMMARY_AGENT_CONFIG_ID,
+          props.data.query.queryString.getQuery().dataset?.dataSource?.id
+        );
+        props.isSummaryAgentAvailable$.next(!!summaryAgentStatus.exists);
+      } catch (error) {
+        console.error(error);
```
Hailong-am, line 61: nit: log.error() or console.error? what's pattern for other api call error handling?

Hailong-am (approved):
would like to have some unit test

public/components/ui_action_context_menu.tsx
Qxisylolo: console.error is more widely used

public/components/ui_action_context_menu.tsx
Qxisylolo: yes, thanks for noticing

common/utils/llm_chat/get_is_data2summary_available.ts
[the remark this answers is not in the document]
Qxisylolo: updated~

Hailong-am:
@Qxisylolo would you mind to fix DCO failure according to https://github.com/opensearch-project/dashboards-assistant/pull/417/checks?check_run_id=37588070203

Qxisylolo:
> @Qxisylolo would you mind to fix DCO failure according to https://github.com/opensearch-project/dashboards-assistant/pull/417/checks?check_run_id=37588070203

Fixed, thanks~

PR #485, opened 2025-03-04
Add error handling when open chatbot and loading converstaion

yyfamazon:
### Description
add error handling when open chatbot and loading converstaion

### Issues Resolved
video 1: test in open source
video 2: clear conversationsError before creating new conversation. 

### Check List

https://github.com/user-attachments/assets/28f80671-11c4-46af-bfc7-b2ec7dc27ddb

https://github.com/user-attachments/assets/4145adb1-3405-480d-b809-33ddbf3db326

### DIFF
--- a/public/tabs/chat/chat_page.tsx
+++ b/public/tabs/chat/chat_page.tsx
@@ -21,9 +21,11 @@ export const ChatPage: React.FC<ChatPageProps> = (props) => {
   const chatContext = useChatContext();
   const { chatState, chatStateDispatch } = useChatState();
   const conversationLoadStatus = useObservable(core.services.conversationLoad.status$);
+  const conversationsStatus = useObservable(core.services.conversations.status$);
   const messagesLoading = conversationLoadStatus === 'loading';
+  const conversationsLoading = conversationsStatus === 'loading';
 
-  const refresh = useCallback(async () => {
+  const refreshConversation = useCallback(async () => {
     if (!chatContext.conversationId) {
       return;
     }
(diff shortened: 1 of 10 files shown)


public/tabs/chat/chat_page.tsx
```diff
@@ -21,9 +21,11 @@ export const ChatPage: React.FC<ChatPageProps> = (props) => {
   const chatContext = useChatContext();
   const { chatState, chatStateDispatch } = useChatState();
   const conversationLoadStatus = useObservable(core.services.conversationLoad.status$);
+  const conversationsStatus = useObservable(core.services.conversations.status$);
```
wanglam, line 24: I'm not prefer to use shared conversation list status here. This status will be used in the conversation list. The chat page will show error if load conversation list failed in that list. 
I have an idea to fix error handling issue here. The conversation load service was designed for chat page. So we can construct new `ConversationsService` in the `ConversationLoadService`. Then we can add a new `loadLatestConversation` method, this new method can be used in `ChatHeaderButton` and `ChatPage`. The `loadLatestConversation` method will use new constructed `ConversationsService` to load conversations. Then it won't effect each other. What do you think about that? @ruanyl

public/tabs/chat/chat_page.tsx
```diff
@@ -39,19 +41,54 @@ export const ChatPage: React.FC<ChatPageProps> = (props) => {
...
               messagesLoading={messagesLoading}
+              conversationsLoading={conversationsLoading}
               messagesLoadingError={
                 typeof conversationLoadStatus !== 'string'
                   ? conversationLoadStatus?.error
                   : undefined
               }
-              onRefresh={refresh}
+              conversationsError={
+                typeof conversationsStatus !== 'string' ? conversationsStatus?.error : undefined
```
wanglam, line 88: When will this `conversationsError` been cleared? Do we need to clear it after creating new conversation?

public/tabs/chat/chat_page.tsx
yyfamazon: after discussion, we will fix this in the future.

public/tabs/chat/chat_page.tsx
wanglam: @yyfamazon What about the new conversation entry from top right text input?

yyfamazon: good point, thanks. pushed a commit and fixed.

wanglam (approved):

PR #493, opened 2025-03-07
Support streaming output

### LINKED ISSUE #387: [WIP][RFC] Support streaming response.
**Is your feature request related to a problem?**
A clear and concise description of what the problem is, e.g. _I'm always frustrated when [...]_

**What solution would you like?**

**What alternatives have you considered?**

**Do you have any additional context?**

SuZhou-Joe:
### Description
This PR is used for supporting streaming output in OpenSearch Chatbot. 

And this PR relies on: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9647 to return a pure http fetch response without the assumption that the content is JSON.

POC video:

https://github.com/user-attachments/assets/d9fe4b2f-21a2-4309-901e-7166abff16ea

This PR choose option B from #387 .

### Issues Resolved
#387 

### Check List

### DIFF
--- a/common/types/chat_saved_object_attributes.ts
+++ b/common/types/chat_saved_object_attributes.ts
@@ -108,4 +108,11 @@ export type StreamChunk =
         messageId: string;
         content: string;
       };
+    }
+  | {
+      event: 'updateOutputMessage';
+      data: {
+        messageId: string;
+        payload: Partial<Omit<IOutput, 'messageId'>>;
+      };
     };
--- a/common/types/chat_saved_object_attributes.ts
+++ b/common/types/chat_saved_object_attributes.ts
@@ -91,10 +91,6 @@ export interface SendResponse {
 
 export type StreamChunk =
   | {
-      event: 'patch';
-      data: Pick<SendResponse, 'interactions' | 'messages'>;
-    }
-  | {
       event: 'error';
       data: string;
     }
--- a/common/types/chat_saved_object_attributes.ts
+++ b/common/types/chat_saved_object_attributes.ts
@@ -99,7 +99,7 @@ export type StreamChunk =
       data: Partial<SendResponse>;
     }
   | {
-      event: 'appendMessage';
+      event: 'appendMessageContent';
       data: {
         messageId: string;
         content: string;
--- a/common/types/chat_saved_object_attributes.ts
+++ b/common/types/chat_saved_object_attributes.ts
@@ -88,3 +88,24 @@ export interface SendResponse {
   messages: IMessage[];
   interactions: Interaction[];
 }
+
+export type StreamChunk =
+  | {
+      event: 'patch';
+      data: Pick<SendResponse, 'interactions' | 'messages'>;
+    }
+  | {
+      event: 'error';
+      data: string;
+    }
+  | {
+      event: 'metadata';
+      data: Partial<SendResponse>;
+    }
+  | {
+      event: 'appendMessage';
+      data: {
+        messageId: string;
+        content: string;
+      };
+    };
--- a/common/utils/stream/serializer.ts
+++ b/common/utils/stream/serializer.ts
@@ -21,7 +21,7 @@ export const streamSerializer = (chunk: StreamChunk): string => {
   return chunkString;
 };
 
-export const sreamDeserializer = (content: string): StreamChunk[] => {
+export const streamDeserializer = (content: string): StreamChunk[] => {
   const streamChunks: StreamChunk[] = [];
   const parser = createParser({
     onEvent(message) {
--- a/common/utils/stream/serializer.ts
+++ b/common/utils/stream/serializer.ts
@@ -1,0 +1,50 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { createParser } from 'eventsource-parser';
+import { StreamChunk } from '../../types/chat_saved_object_attributes';
+
+const separators = `\n\n`;
+const prefix = 'data: ';
+
+export const streamSerializer = (chunk: StreamChunk): string => {
+  const { event, data } = chunk;
+  let chunkString = '';
+  if (event) {
+    chunkString += `event: ${event}\n`;
+  }
+
+  chunkString += `${prefix}${JSON.stringify(data)}${separators}`;
+
+  return chunkString;
+};
+
+export const sreamDeserializer = (content: string): StreamChunk[] => {
+  const streamChunks: StreamChunk[] = [];
+  const parser = createParser({
+    onEvent(message) {
+      try {
+        const dataPayload = JSON.parse(message.data);
+        streamChunks.push({
+          ...message,
+          data: dataPayload,
+        } as StreamChunk);
+      } catch (e) {
+        streamChunks.push({
+          event: 'error',
+          data: e.message,
+        });
+      }
+    },
+    onError(error) {
+      streamChunks.push({
+        event: 'error',
+        data: error.message,
+      });
+    },
+  });
+  parser.feed(content);
+  return streamChunks;
+};
--- a/public/components/blink_cursor.scss
+++ b/public/components/blink_cursor.scss
@@ -1,5 +1,5 @@
 .assistant_blinkCursor {
-  border-right: 2px solid black;
+  border-right: 2px solid $euiColorDarkShade;
   animation: cursor-blink 0.8s infinite;
 }
 
--- a/public/components/blink_cursor.scss
+++ b/public/components/blink_cursor.scss
@@ -1,0 +1,8 @@
+.assistant_blinkCursor {
+  border-right: 2px solid black;
+  animation: cursor-blink 0.8s infinite;
+}
+
+@keyframes cursor-blink {
+  50% { border-right-color: transparent; }
+}
--- a/public/components/markdown_with_blink_cursor.tsx
+++ b/public/components/markdown_with_blink_cursor.tsx
@@ -1,0 +1,48 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { EuiMarkdownFormat, EuiMarkdownFormatProps } from '@elastic/eui';
+import React from 'react';
+import { useRef } from 'react';
+import { useEffect } from 'react';
+
+function getLastTextNode(node: HTMLElement | Node): Node | null | undefined {
+  if (node.nodeType === node.TEXT_NODE) {
+    return node;
+  }
+  const children = node.childNodes;
+  for (let i = children.length - 1; i >= 0; i--) {
+    const child = children[i];
+    const result = getLastTextNode(child);
+    if (result) {
+      return result;
+    }
+    return null;
+  }
+}
+
+export const MarkdownWithBlinkCursor = (props: EuiMarkdownFormatProps & { loading?: boolean }) => {
+  const ref = useRef<HTMLDivElement | null>(null);
+  useEffect(() => {
+    let blinkCursorNode: undefined | HTMLSpanElement;
+    if (props.loading && ref.current) {
+      const lastNode = getLastTextNode(ref.current);
+      blinkCursorNode = document.createElement('span');
+      blinkCursorNode.classList.add('assistant_blinkCursor');
+      if (lastNode) {
+        lastNode.parentNode?.appendChild(blinkCursorNode);
+      }
+    }
+
+    return () => {
+      blinkCursorNode?.remove();
+    };
+  }, [props.children, props.loading]);
+  return (
+    <div className="markdown_with_blink_cursor" ref={ref}>
+      <EuiMarkdownFormat>{props.children}</EuiMarkdownFormat>
+    </div>
+  );
+};
--- a/public/components/markdown_with_blink_cursor.tsx
+++ b/public/components/markdown_with_blink_cursor.tsx
@@ -24,7 +24,7 @@ function getLastTextNode(node: HTMLElement | Node): Node | null | undefined {
 }
 
 export const MarkdownWithBlinkCursor = (props: EuiMarkdownFormatProps & { loading?: boolean }) => {
-  const ref = useRef<HTMLDivElement | null>(null);
+  const ref = useRef<HTMLDivElement>(null);
   useEffect(() => {
     let blinkCursorNode: undefined | HTMLSpanElement;
     if (props.loading && ref.current) {
--- a/public/components/markdown_with_blink_cursor.tsx
+++ b/public/components/markdown_with_blink_cursor.tsx
@@ -4,23 +4,93 @@
  */
 
 import { EuiMarkdownFormat, EuiMarkdownFormatProps } from '@elastic/eui';
-import React from 'react';
-import { useRef } from 'react';
-import { useEffect } from 'react';
+import React, { useRef, useEffect } from 'react';
 
-function getLastTextNode(node: HTMLElement | Node): Node | null | undefined {
-  if (node.nodeType === node.TEXT_NODE) {
+const voidElements = [
+  'area',
+  'base',
+  'br',
+  'col',
+  'embed',
+  'hr',
+  'img',
+  'input',
+  'link',
+  'meta',
+  'param',
+  'source',
+  'track',
+  'wbr',
+];
+
+function isVoidElements(tagName: string | null) {
+  if (!tagName) {
+    return true;
+  }
+
+  return voidElements.includes(tagName.toLowerCase());
+}
+
+function getTagName(node: Node | HTMLElement): string | null {
+  // Check if the node is an element node
+  if (node.nodeType === Node.ELEMENT_NODE) {
+    // It's safe to cast to Element and access tagName
+    return (node as Element).tagName.toLowerCase();
+  }
+
+  // For non-element nodes (text nodes, comment nodes, etc.)
+  return null;
+}
+
+/**
+ * Find most right leaf node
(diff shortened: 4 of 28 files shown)


public/components/markdown_with_blink_cursor.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const children = node.childNodes;
+  for (let i = children.length - 1; i >= 0; i--) {
+    const child = children[i];
+    const result = getLastTextNode(child);
+    if (result) {
+      return result;
+    }
+    return null;
+  }
+}
```
FriedhelmWS, line 24: How about
```
function isTextNode(node: Node): node is Text {
  return node.nodeType === Node.TEXT_NODE;
}

function getLastTextNode(node: HTMLElement | Node): Node | null | undefined {
  if (isTextNode(node)) {
    return node;
  }
  const children = node.childNodes;
  for (let i = children.length - 1; i >= 0; i--) {
    const child = children[i];
    if (isTextNode(child) && (child as Text).data === '\n') {
      continue;
    }
    const result = getLastTextNode(child);
    if (result) {
      return result;
    }
    return null;
  }
}
```
So the blink mark doesn't append to the next line when the current node is `li` element.

public/hooks/use_chat_state.tsx
```diff
@@ -122,9 +147,31 @@ const chatStateReducer: React.Reducer<ChatState, ChatStateAction> = (state, acti
...
+          const patchMessages: IMessage[] = [
+            {
+              ...updatingMessage,
+              content: updatingMessage.content + action.payload.content,
+            },
+          ];
+          draft.messages = addPatchInArray(state.messages, patchMessages, 'messageId');
+        }
+        break;
+
```
FriedhelmWS, line 167: How about declare `apepndMessage` like this:
```
{
      type: 'appendMessage';
      payload: {
        content: string;
        messageId: string;
        suggestedActions?: ISuggestedAction[];
      };
    };
```
and modify there to be something like:
```
          const patchMessages: IMessage[] = [
            {
              ...updatingMessage,
              content: updatingMessage.content + action.payload.content,
              ...(action.payload.suggestedActions && {
                suggestedActions: [
                  ...((updatingMessage as IOutput).suggestedActions || []),
                  ...action.payload.suggestedActions,
                ],
              }),
            },
          ];
```
so we can also handle the appending of suggested actions.

SuZhou-Joe: I am thinking we should add a new action like
```
{
  event: 'updateOutputMessage',
  data: {
    messageId: string;
    payload: Partial<IOutput>
  }
}
```
to keep appendMessage clean.

public/components/markdown_with_blink_cursor.tsx
SuZhou-Joe: We can have a follow up PR for this enhancement. This PR is a little bit large.

public/components/blink_cursor.scss
```diff
@@ -0,0 +1,8 @@
+.assistant_blinkCursor {
+  border-right: 2px solid black;
```
SuZhou-Joe, line 2: Need to comply with OUI styling.

public/components/markdown_with_blink_cursor.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const MarkdownWithBlinkCursor = (props: EuiMarkdownFormatProps & { loading?: boolean }) => {
+  const ref = useRef<HTMLDivElement | null>(null);
```
FriedhelmWS, line 27: Minor: I think `| null` here is unnecessary since HTMLDivElement already includes type `null`.

common/types/chat_saved_object_attributes.ts
```diff
@@ -88,3 +88,31 @@ export interface SendResponse {
...
+    }
+  | {
+      event: 'appendMessage';
+      data: {
+        messageId: string;
+        content: string;
+      };
+    }
+  | {
+      event: 'updateOutputMessage';
```
FriedhelmWS, line 113: Minor: would the name of this even slightly confusing? As this event can actually update not only message content but all other values in `IOutput` so maybe something like `updateChunk` would be more appropriate?

SuZhou-Joe: IMessage is a union type of IInput and IOutput, OutputMessage means message whose type is output actually. And this action is used for updating metadata of outputMessage.

`updateChunk` to me seems to be able to describe any type of chunk from backend.

public/components/markdown_with_blink_cursor.tsx
SuZhou-Joe: Sure, updated.

public/components/markdown_with_blink_cursor.tsx
```diff
@@ -0,0 +1,48 @@
...
+    return node;
+  }
+  const children = node.childNodes;
+  for (let i = children.length - 1; i >= 0; i--) {
+    const child = children[i];
+    const result = getLastTextNode(child);
+    if (result) {
+      return result;
+    }
+    return null;
```
yubonluo, line 22: Should we move `return null` to the for loop external? This causes the `getLastTextNode` to return immediately after checking the last child node.

public/utils/message_content_pool.ts
```diff
@@ -0,0 +1,94 @@
...
+      // Update the content pool with remaining content
+      this.messageContentChunk$.next(restContents);
+
+      if (this.inputCompleted && !Object.keys(restContents).length) {
+        this.stop();
+      } else {
+        // Schedule the next processing cycle
+        this.startMessageContentJob();
+      }
+    }, JOB_INTERVAL);
```
raintygao, line 56: Is `JOB_INTERVAL = 50` coming from a Intuition or some calculation?

public/utils/message_content_pool.ts
```diff
@@ -0,0 +1,94 @@
...
+import { BehaviorSubject, Observable, Subscriber } from 'rxjs';
+
+const JOB_INTERVAL = 50;
+const CONTENT_SLICE_LENGTH = 10;
+
+export class MessageContentPool {
+  // Backend may give large response in a single chunk,
+  // use an observable here to buffer the large response.
+  private messageContentChunk$: BehaviorSubject<Record<string, string>>;
+  private messageContentTimer: NodeJS.Timeout | null = null;
```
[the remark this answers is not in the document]
raintygao: I guess it may be OK because data plugin does the same thing, and @types/node is declared as a devDependency of OSD.

server/routes/chat_routes.ts
```diff
@@ -198,6 +198,19 @@ export function registerChatRoutes(router: IRouter, routeOptions: RoutesOptions)
         return response.custom({ statusCode: error.statusCode || 500, body: error.message });
       }
 
+      if (outputs.stream) {
+        const result = response.ok({
+          headers: {
+            'Content-Encoding': 'identity',
```
FriedhelmWS, line 204: Minor: should we add some explanation for the reason of this?

public/utils/message_content_pool.ts
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { BehaviorSubject, Observable, Subscriber } from 'rxjs';
+
+const JOB_INTERVAL = 50;
+const CONTENT_SLICE_LENGTH = 10;
```
FriedhelmWS, line 9: Should we make the actual streaming speed configurable?

public/hooks/use_get_chunks_from_http_response.tsx
```diff
@@ -0,0 +1,162 @@
...
+      const response = (await props.fetchResponse.body) as Partial<SendResponse>;
+      chunk$.next({
+        event: 'metadata',
+        data: response,
+      });
+
+      // Complete right after next will only eat the emit value
+      // add a setTimeout to delay it into next event loop
+      setTimeout(() => {
+        messageContentPool.inputComplete();
```
wanglam, line 92: Why we need to call `messageContentPool.inputComplete`  here? For pure text response, seems we don't need to append message to the message pool.

public/hooks/use_chat_actions.tsx
```diff
@@ -3,83 +3,116 @@
...
-          messages: response.messages,
-          interactions: response.interactions,
-        },
-      });
-    } catch (error) {
-      if (abortController.signal.aborted) return;
-      chatStateDispatch({ type: 'error', payload: error });
-    }
-  };
+      chunk$.subscribe(
```
wanglam, line 58: Combine the text response and the streaming response in the same observable is a little bit complicated to me. How about we check the response type outside and use the JSON response directly?

public/components/markdown_with_blink_cursor.tsx
```diff
@@ -0,0 +1,48 @@
...
+      const lastNode = getLastTextNode(ref.current);
+      blinkCursorNode = document.createElement('span');
+      blinkCursorNode.classList.add('assistant_blinkCursor');
+      if (lastNode) {
+        lastNode.parentNode?.appendChild(blinkCursorNode);
+      }
+    }
+
+    return () => {
+      blinkCursorNode?.remove();
```
wanglam, line 40: Seems this `blinkCursorNode` won't be removed after message generated completed. There are multi cursors during new message streaming output.

public/utils/message_content_pool.ts
```diff
@@ -0,0 +1,94 @@
...
+            messageContent: messageContent.slice(0, CONTENT_SLICE_LENGTH),
+          });
+          const restContent = messageContent.slice(CONTENT_SLICE_LENGTH);
+          if (restContent.length > 0) {
+            return {
+              ...acc,
+              [messageId]: restContent,
+            };
+          }
+          return acc;
```
wanglam, line 42: Do we need to clear current content if `resetContent.length` equal 0? 
For example, the value of `messageContentChunk$`
```
{"message-1":"foo"}
```
The `outputSubscriber` will emit '{messageId:"message-1", messageContent: "foo"}'.
Since the restContent.length equal zero, it will return original `acc`. The messageContent for "message-1" still "foo".

public/utils/message_content_pool.ts
```diff
@@ -0,0 +1,94 @@
...
+        },
+        {}
+      );
+
+      // Update the content pool with remaining content
+      this.messageContentChunk$.next(restContents);
+
+      if (this.inputCompleted && !Object.keys(restContents).length) {
```
wanglam, line 50: When will the `Object.keys(restContents)` equal 0? Seems messageId won't be deleted after calling `addMessageContent`.

public/components/markdown_with_blink_cursor.tsx
SuZhou-Joe: Yeah, I refactored this part, please revisit.

server/services/chat/chat_service.ts
```diff
@@ -14,6 +15,7 @@ export interface ChatService {
     messages: IMessage[];
     conversationId: string;
     interactionId: string;
+    stream?: Stream; // This is the stream to update the message
```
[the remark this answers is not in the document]
SuZhou-Joe: As agent framework from ML plugin is JSON based so you won't find code consuming this field. But there is a POC for this: https://github.com/SuZhou-Joe/dashboards-assistant/pull/1/files#diff-c20d6355af792c9cdb22f0f87954d400a8cd0ce0ad3c67efec7bf216947d1eb8R120 .

public/utils/message_content_pool.ts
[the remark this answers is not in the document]
SuZhou-Joe: Changed to `ReturnType<typeof setTimeout>` to keep it campatible for both node env and browser env.

public/utils/message_content_pool.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { BehaviorSubject, Observable, Subscriber } from 'rxjs';
+
+const JOB_INTERVAL = 50;
+const CONTENT_SLICE_LENGTH = 10;
+
+export class MessageContentPool {
```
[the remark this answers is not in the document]
SuZhou-Joe: I am fine with that, updated.

public/utils/message_content_pool.ts
```diff
@@ -0,0 +1,94 @@
...
+
+      // Update the content pool with remaining content
+      this.messageContentChunk$.next(restContents);
+
+      if (this.inputCompleted && !Object.keys(restContents).length) {
+        this.stop();
+      } else {
+        // Schedule the next processing cycle
+        this.startMessageContentJob();
+      }
```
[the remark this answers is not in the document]
SuZhou-Joe: Yeah, actually if any error happens the job will stop. But it does no harm to add error handling in case the logic becomes more complicated in the future.

server/routes/chat_routes.ts
SuZhou-Joe: Sure, I added some comment in the fetch module but we should add comment in here as well.

public/utils/message_content_pool.ts
SuZhou-Joe: Actually Intuition from practice, any advice for the speed?

public/utils/message_content_pool.ts
SuZhou-Joe: Yeah, I was trying to but can not find a use case for that. Updated anyway, dev for the future.

common/utils/stream/serializer.ts
```diff
@@ -0,0 +1,50 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { createParser } from 'eventsource-parser';
+import { StreamChunk } from '../../types/chat_saved_object_attributes';
+
+const separators = `\n\n`;
+const prefix = 'data: ';
+
+export const streamSerializer = (chunk: StreamChunk): string => {
```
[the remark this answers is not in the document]
SuZhou-Joe: I chose not to use `eventsource-encoder` here because:

1. The `Weekly Downloads` for this package is much smaller than the decoder, I am not confident to use it.
2. Encoder is much more simple than Decode.

public/hooks/use_chat_state.tsx
```diff
@@ -38,7 +44,33 @@ type ChatStateAction =
...
+      };
+    }
+  | {
+      type: 'updateResponseType';
+      payload: {
+        type: LLMResponseType;
+      };
+    }
+  | {
+      type: 'appendMessage';
```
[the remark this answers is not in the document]
SuZhou-Joe: Sure, updated.

public/components/markdown_with_blink_cursor.tsx
SuZhou-Joe: the useEffect listen to `[props.children, props.loading]`, it will remove previous node if content get changed.

server/services/chat/chat_service.ts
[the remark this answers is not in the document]
Hailong-am: make sense, could we have a video with your POC?

public/hooks/use_chat_actions.tsx
SuZhou-Joe: Make sense, the hook is a little bit complicated now.

public/hooks/use_get_chunks_from_http_response.tsx
SuZhou-Joe: Removed since we refactor this hook.

public/utils/message_content_pool.ts
raintygao: Not any others, just out of curious, this value looks good to me.

SuZhou-Joe: Get it, added some comment to reduce confusion.

server/services/chat/chat_service.ts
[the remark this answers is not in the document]
SuZhou-Joe: Sure, updated.

public/components/markdown_with_blink_cursor.tsx
SuZhou-Joe: Refactor the node and address the issue, please take a look.

Hailong-am (approved):

public/hooks/use_get_chunks_from_http_response.tsx
```diff
@@ -0,0 +1,123 @@
...
+              messageContentPuller.addMessageContent(
+                chunkObject.data.messageId,
+                chunkObject.data.content
+              );
+            } else {
+              chunk$.next(chunkObject);
+            }
+          }
+        } catch (e) {
+          chunk$.error(e);
```
wanglam, line 48: Nit: Do we need to call `messageContentPuller.inputComplete();` or `messageContentPuller.stop();` here?

public/hooks/use_get_chunks_from_http_response.tsx
```diff
@@ -0,0 +1,123 @@
...
+          return;
+        }
+      },
```
wanglam, line 51: Nit: Do we need to add error subscribe here?

wanglam (approved):

raintygao (approved):

public/hooks/use_get_chunks_from_http_response.tsx
SuZhou-Joe: Sure, updated.

public/hooks/use_get_chunks_from_http_response.tsx
SuZhou-Joe: Sure, updated.

PR #501, opened 2025-03-12
Adjust the style of alert summary popover

FriedhelmWS:
### Description
1. Change the insight with RAG button from icon button to text button
2. Allow the user to back to the alert summary after viewing the insight with RAG 
3. Change the text of discover button and move it to be inline with other buttons
4. Increase the overall padding for the summary context
Screenshot 2025-03-12 at 13 54 27
Screenshot 2025-03-12 at 13 53 53
Screenshot 2025-03-12 at 13 53 46

### Issues Resolved
NA

### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass, including unit test, integration test.
- [ ] New functionality has user manual doc added.
- [ ] Commits are signed per the DCO using --signoff.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #506, opened 2025-03-14
Change the background color, button position and text for alert summary popover

FriedhelmWS:
### Description
Change the alert summary popover layout and button text:
Screenshot 2025-03-17 at 09 56 31
Dark mode:
Screenshot 2025-03-17 at 09 54 09

Please notice this PR contains some of the changes that revert the previous PR https://github.com/opensearch-project/dashboards-assistant/pull/501

This is because, after further consideration, I think continuing to reuse the info icon button (originally used for 'how was it generated' inside the chatbot) is a bad idea. This results in a single button being used for two completely different purposes, requiring many conditional checks, which makes the code bulky.
Screenshot 2025-03-14 at 12 17 00

Therefore, in this PR, the 'view insight' button has been completely separated from the message action group and is now a standalone button. Now we always hide the info button icon for alert summary popover and conditionally display the 'view insight' button outside of message action button group.

### Issues Resolved
NA

### Check List

### DIFF
--- a/public/components/incontext_insight/generate_popover_body.tsx
+++ b/public/components/incontext_insight/generate_popover_body.tsx
@@ -18,6 +18,7 @@ import {
   EuiText,
   EuiTitle,
   EuiButton,
+  EuiButtonEmpty,
 } from '@elastic/eui';
 import { useEffectOnce } from 'react-use';
 import { METRIC_TYPE } from '@osd/analytics';
@@ -263,12 +264,12 @@ export const GeneratePopoverBody: React.FC<{
     const content = showInsight && insightAvailable ? insight : summary;
     return content ? (
       <>
-        <EuiPanel paddingSize="s" hasBorder hasShadow={false} color="subdued">
+        <EuiPanel paddingSize="s" hasBorder hasShadow={false}>
           <EuiText className="incontextInsightGeneratePopoverContent" size="s">
             <EuiMarkdownFormat>{content}</EuiMarkdownFormat>
           </EuiText>
           <EuiSpacer size={'xs'} />
-          {renderInnerFooter()}
+          {renderInnerFooter(content)}
         </EuiPanel>
       </>
     ) : (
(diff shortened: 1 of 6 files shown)


wanglam:
@FriedhelmWS Could you added an item in the `release-notes/dashboards-assistant.release-notes-3.0.0.0-alpha1.md` if this PR target to `3.0.0.0-alpha1`?

public/tabs/chat/messages/message_action.tsx
```diff
@@ -171,23 +167,18 @@ export const MessageActions: React.FC<MessageActionsProps> = ({
     },
     trace: {
```
wanglam, line 168: Nit: Since this trace button won't be used anymore. why don't we remove this button config?

public/components/incontext_insight/generate_popover_body.tsx
```diff
@@ -338,12 +357,7 @@ export const GeneratePopoverBody: React.FC<{
             <MessageActions
               contentToCopy={summary}
```
wanglam, line 358: Nit: Not related to this PR, it seems the only difference between these two `MessageActions` renders is `contentToCopy`. To keep the copied content the same as what is displayed, I'm wondering if we can add an option in `renderInnerFooter`. This way, we wouldn't need to check which message actions to be displayed.

public/tabs/chat/messages/message_action.tsx
FriedhelmWS: I think the tract button is still used on the message bubble for chatbot
Screenshot 2025-03-17 at 11 34 02

public/components/incontext_insight/generate_popover_body.tsx
FriedhelmWS: Done, thanks for the suggestion so that we can remove the duplicate code

wanglam (approved):

PR #514, opened 2025-03-20
Add auto aggregation suggest for t2v

Hailong-am:
### Description

Add auto aggregation suggest for t2v. It will depends on whether index pattern has time field or not, if it don't have time field, will simply use `count` as aggregation method, otherwise, trying to figure out the best interval for date histogram

### Issues Resolved

### Check List

### DIFF
--- a/public/utils/pipeline/ppl_aggs_auto_suggest_task.ts
+++ b/public/utils/pipeline/ppl_aggs_auto_suggest_task.ts
@@ -1,0 +1,73 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Task } from './task';
+import { DataPublicPluginStart } from '../../../../../src/plugins/data/public';
+
+interface Input {
+  ppl: string;
+  dataSourceId: string | undefined;
+  timeFiledName?: string;
+}
+
+export class PPLAutoSuggestTask extends Task<Input, Input> {
+  searchClient: DataPublicPluginStart['search'];
+
+  constructor(searchClient: DataPublicPluginStart['search']) {
+    super();
+    this.searchClient = searchClient;
+  }
+
+  async execute<T extends Input>(v: T) {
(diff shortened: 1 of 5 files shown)


public/utils/pipeline/ppl_aggs_auto_suggest_task.ts
```diff
@@ -0,0 +1,73 @@
...
+                { strategy: 'pplraw' }
+              )
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              .toPromise<any>();
+
+            if (res.rawResponse.total === 0) {
+              ppl = `${ppl} | stats count()`;
+            } else {
+              // get min and max and calculate proper interval by range
+              const min = res.rawResponse.jsonData[0].min;
```
gaobinlong, line 47: If the generated ppl contains time range condition, then should we use that condition directly to narrow down the time range? Too many buckets maybe created if we use the max/min time in the index, which has poor performance.

Hailong-am: we didn't remove any filter in original ppl, what we do is get the range for existing query result and append a date histogram aggregation

public/utils/pipeline/ppl_aggs_auto_suggest_task.ts
```diff
@@ -0,0 +1,73 @@
...
+          try {
+            res = await this.searchClient
+              .search(
+                { params: { body: { query: dateRangePPL } }, dataSourceId: v.dataSourceId },
+                { strategy: 'pplraw' }
+              )
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              .toPromise<any>();
+
+            if (res.rawResponse.total === 0) {
```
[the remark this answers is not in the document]
Hailong-am: Yes, let me add the null check

ruanyl (approved):

wanglam (approved):

PR #539, opened 2025-04-14
style single metric in text2vis

Qxisylolo:
### Description
This pr styles the single metric by adding a text description layer

### Issues Resolved
before:
截屏2025-04-14 11 26 45

after:
截屏2025-04-14 11 27 53

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Enhancements
 
+- Style single metric in text2vis ([#539](https://github.com/opensearch-project/dashboards-assistant/pull/539))
+
 ### Bug Fixes
 
 ### Infrastructure
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,6 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Enhancements
 
-- Style single metric in text2vis ([#539](https://github.com/opensearch-project/dashboards-assistant/pull/539))
-
 ### Bug Fixes
 
 ### Infrastructure
--- a/server/utils/style_single_metric.ts
+++ b/server/utils/style_single_metric.ts
@@ -3,7 +3,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-export function addTitleTextLayer(json: any) {
+interface VegaLiteSpec {
+  title?: string;
+  mark?: unknown;
+  encoding?: unknown;
+  layer?: object[];
+}
+
+export function addTitleTextLayer(json: VegaLiteSpec) {
   if (!json.title) return json;
 
   const titleTextLayer = {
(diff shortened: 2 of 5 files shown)


server/utils/style_single_metric.ts
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export function addTitleTextLayer(json: any) {
```
raintygao, line 6: Could we avoid using `any` here?

server/utils/style_single_metric.ts
```diff
@@ -0,0 +1,58 @@
...
+            type: 'text',
+            fontSize: 80,
+            fontWeight: 'bold',
+          },
+          encoding,
+        },
+      ],
+    };
+  }
+  layeredSpec.layer.push(titleTextLayer);
```
raintygao, line 41: Nit: Why do we hard code a layer and push a fixed layer seperately?

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Enhancements
 
+- Style single metric in text2vis ([#539](https://github.com/opensearch-project/dashboards-assistant/pull/539))
+
```
raintygao, line 12: We have deprecated changelog, please track your changes in release note.

server/utils/style_single_metric.ts
Qxisylolo: thanks for the comment, normally, we won't have a layer in output. and the output is usually something like this:
`<vega-lite>
{
  "title": "Total Data Count",
  "description": "The total number of data points",
  "mark": "text",
  "encoding": {
    "text": {
      "field": "count()",
      "type": "quantitative",
      "axis": {
        "title": "count()"
      }
    }
  }
}
</vega-lite>`
in this case, we shall  manually create a layer array (in vega, to display 2 lines of mark, we shall use layer)and push the titleTextLayer into it.

server/utils/style_single_metric.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  return layeredSpec;
+}
+
+export function checkSingleMetric(textContent: String) {
+  const metricsMatch = textContent.match(
+    /Number of metrics:\s*\[[^\]]*\]\s*<number of metrics\s*\{(\d+)\}>/
```
ruanyl, line 48: Can we check the respond vega-lite json to see if it's a single text visualization? Then we don't need to rely on specific text contents returned by LLM.

Hailong-am: +1, instead we should rely on ppl or sampleData, for single value case, sample data will only have 1 row and 1 column

CHANGELOG.md
Qxisylolo: fixed

server/utils/style_single_metric.ts
Qxisylolo: updated, we will use sampleData to check if it is single value.

server/utils/style_single_metric.ts
Qxisylolo: updated

yubonluo (approved):

Hailong-am (approved):

ruanyl (approved):

PR #542, opened 2025-04-14
Fix .plugins-ml-memory-meta not found when get conversations

wanglam:
### Description
This PR is to fix the conversation loading error when `.plugins-ml-memory-meta` index not exists like screenshot below. 
image
It will return an empty result after this PR merged.

### Issues Resolved

### Check List

### DIFF
--- a/server/routes/chat_routes.ts
+++ b/server/routes/chat_routes.ts
@@ -290,6 +290,14 @@ export function registerChatRoutes(router: IRouter, routeOptions: RoutesOptions)
         const getResponse = await storageService.getConversations(request.query);
         return response.ok({ body: getResponse });
       } catch (error) {
+        const errorBody = error?.meta?.body?.error;
+        // Ignore index not found exception for .plugins-ml-memory-meta index
+        if (
+          errorBody?.type === 'index_not_found_exception' &&
+          errorBody?.reason.includes('.plugins-ml-memory-meta')
+        ) {
+          return response.ok({ body: { objects: [], total: 0 } });
+        }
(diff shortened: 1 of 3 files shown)


server/routes/chat_routes.ts
```diff
@@ -290,6 +290,14 @@ export function registerChatRoutes(router: IRouter, routeOptions: RoutesOptions)
         const getResponse = await storageService.getConversations(request.query);
         return response.ok({ body: getResponse });
       } catch (error) {
+        const errorBody = error?.meta?.body?.error;
+        // Ignore index not found exception for .plugins-ml-memory-meta index
+        if (
+          errorBody?.type === 'index_not_found_exception' &&
+          errorBody?.reason.includes('.plugins-ml-memory-meta')
```
Hailong-am, line 297: we can use status code 404? hardcode index name here is not a good practice, index name may changes in future.

wanglam: Sure, it can be changed to a 404 status code to determine if we need to return an empty result. The reason for using `plugins-ml-memory-meta` here is that I'm not sure if any other 404 errors will be thrown here. What do you think about this?

Hailong-am: I think use status code 404 should be safe, considering support aoss or multi tenancy, we don't have the assumption that the backend storage is OpenSearch index.

wanglam: Updated, could you help me take a look?

Hailong-am (approved):
LGTM

raintygao (approved):

PR #549, opened 2025-04-18
[Enhancement] Buffer for special characters when streaming

SuZhou-Joe:
### Description

When the streaming feature meets hyperlink, the experience becomes weird that the underlying link will be visible at first chunks and becomes invisible after following chunks come in.

#### Before fix

https://github.com/user-attachments/assets/aeb90afc-7b89-4381-becb-6522882fcf41

#### After fix

https://github.com/user-attachments/assets/2f5ffd42-e192-47a8-99c3-8b5ef9cbfe4b

### Issues Resolved

### Check List

### DIFF
--- a/public/hooks/use_get_chunks_from_http_response.tsx
+++ b/public/hooks/use_get_chunks_from_http_response.tsx
@@ -20,10 +20,10 @@ export const useGetChunksFromHTTPResponse = () => {
     const chunk$ = new BehaviorSubject<StreamChunk | undefined>(undefined);
     const messageContentPuller = new MessageContentPuller({
       isContentReadyToUse(message) {
-        const regexpForHyberlink = /\[([^\]\n]+)\]\(([^)\n]+)\)/;
+        const regexpForHyperlink = /\[([^\]\n]+)\]\(([^)\n]+)\)/;
 
-        // For message contains [, but not completely comply with hyberlink format
-        if (message.includes('[') && !regexpForHyberlink.test(message)) {
+        // For message contains [, but not completely comply with hyperlink format
+        if (message.includes('[') && !regexpForHyperlink.test(message)) {
           return false;
         }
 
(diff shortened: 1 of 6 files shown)


FriedhelmWS (approved):
LGTM

public/utils/message_content_puller.test.ts
```diff
@@ -21,21 +23,78 @@ describe('MessageContentPool', () => {
...
+          messageId: 'b',
+        });
+        expect(subscriptionMock).toHaveBeenNthCalledWith(3, {
+          messageContent: 'a'.repeat(10),
+          messageId: 'a',
+        });
+        expect(completeSubscriptionMock).toHaveBeenCalledTimes(1);
+      },
+      {
+        timeout: 10000,
```
wanglam, line 44: Nit: can we use `jest.useFakeTimer` here?  Then these tests won't take too much time.

wanglam (approved):

public/hooks/use_get_chunks_from_http_response.tsx
```diff
@@ -18,7 +18,18 @@ export const useGetChunksFromHTTPResponse = () => {
...
+        const regexpForHyperlink = /\[([^\]\n]+)\]\(([^)\n]+)\)/;
+
+        // For message contains [, but not completely comply with hyperlink format
+        if (message.includes('[') && !regexpForHyperlink.test(message)) {
+          return false;
+        }
+
+        return true;
+      },
+    });
```
Hailong-am, line 32: It's nice we have this logic to improve rendering experience. My concern is markdown itself support lots of format like table, it will render differently once it reached to its end tag. I would prefer not add those special logic.

Also, if the output stoped at the middle of a hyperlink due to max output token limit reached, what will displayed to user?

Hailong-am (approved):

public/hooks/use_get_chunks_from_http_response.tsx
SuZhou-Joe: In that case, we will display [hyperlink](http://www.link-not-closed.

> I would prefer not add those kind of special logics.
Yeah, the hyperlink experience is bad so we have to address it. Actually I tried for table format, and that works fine without properly closed. 
image

public/utils/message_content_puller.test.ts
SuZhou-Joe: Sure, updated.

PR #553, opened 2025-04-24
Save chatbot flyout visualize state to local storage

wanglam:
### Description

This PR save chat flyout visualize state in user browser local storage. The chatbot flyout will keep open when user navigate to other pages after this PR get merged.

See videos below:

https://github.com/user-attachments/assets/e80a87d1-4974-4e6d-954d-7163167ae9e0

### Issues Resolved

### Check List

### DIFF
--- a/public/chat_header_button.tsx
+++ b/public/chat_header_button.tsx
@@ -5,9 +5,15 @@
 
 import { EuiButtonIcon, EuiToolTip } from '@elastic/eui';
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
-import { useEffectOnce, useObservable } from 'react-use';
+import { useEffectOnce, useObservable, useUpdateEffect } from 'react-use';
+import { debounceTime } from 'rxjs/operators';
 
-import { ApplicationStart, HeaderVariant, SIDECAR_DOCKED_MODE } from '../../../src/core/public';
+import {
+  ApplicationStart,
+  HeaderVariant,
+  MountPoint,
+  SIDECAR_DOCKED_MODE,
+} from '../../../src/core/public';
 import { getIncontextInsightRegistry } from './services';
 import { ChatFlyout } from './chat_flyout';
 import { ChatContext, IChatContext } from './contexts/chat_context';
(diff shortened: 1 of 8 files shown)


public/contexts/chat_header_button_store.ts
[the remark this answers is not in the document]
wanglam: Should be removed.

ruanyl (approved):

public/services/index.ts
```diff
@@ -43,6 +44,8 @@ export const [getAssistantService, setAssistantService] = createGetterSetter<Ass
   'AssistantServiceStart'
 );
 
+export const [getLocalStorage, setLocalStorage] = createGetterSetter<DataStorage>('LocalStorage');
+
```
raintygao, line 48: Is there any special intention that we memorize a `localstorage` here?  In my intuition we can use `window.localStorage` anywhere.

public/chat_header_button.tsx
```diff
@@ -36,33 +48,86 @@ interface HeaderChatButtonProps {
...
+      .getSidecarConfig$()
+      .pipe(debounceTime(30))
+      .subscribe((newSidecarConfig) => {
+        if (newSidecarConfig) {
+          setChatbotSidecarConfig(newSidecarConfig);
+        }
+      });
+    let timeoutId: number;
+    if (flyoutVisible) {
+      // Add setTimeout here to avoid "container was removed without using React." error
```
raintygao, line 118: Would this be a warning or error?

public/chat_header_button.tsx
```diff
@@ -36,33 +48,86 @@ interface HeaderChatButtonProps {
...
+      timeoutId = window.setTimeout(() => {
+        if (flyoutMountPoint.current) {
+          loadLatestConversation();
+          openSidecar(flyoutMountPoint.current);
+        }
+      }, 0);
```
raintygao, line 124: Nit: Im thinking if `requestAnimationFrame` here would be better here.

public/chat_header_button.tsx
wanglam: image
It's a warning. But the chatbot can't be rendered after that. I guess it may related to the `cleanupDom` function in the sidecar service:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/overlays/sidecar/sidecar_service.tsx#L194

The chatbot flyout can be rendered without `setTimeout` after that. But still got some warnings like screenshot below after page switching:
image
 
So I used `setTimeout` here to avoid these warning messages.

public/services/index.ts
wanglam: There are two benefits of using `LocalStorage` here:
1. All storage items can be stored under the same namespace ("dashboardsAssistant."). It's better to manage storage items under the dashboards-assistant plugin.
2. No need to manually `JSON.stringify` or `JSON.parse` the storage items. They will be converted by [DataStorage](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/common/storage/storage.ts).

raintygao (approved):

Hailong-am (approved):

